PR #27, opened 2025-01-29
Bedrock basic answer strategy

chaecramb:
This PR implements a basic answer strategy using Claude 3.5 Sonnet on Bedrock. In this initial implementation we simply have Claude provide a structured JSON response directly, without rephrasing, guardrails or RAG. 

It does four things:

- Allow configuration of answer strategy via environment variables.
- Convert the `OpenAIAnswer` class to a more generic `PipelineRunner` class, this can be used to run the pipeline for both Claude and OpenAI.
- Creates a `StructuredAnswerComposer` for Claude that returns a JSON response via a tool call.
- Sets up and implements initial unit and system tests for Claude, along with relevant helpers.

Trello: https://trello.com/c/wB5dUyQ7/2257-integrate-a-very-basic-answer-strategy-no-rag-with-aws-bedrock

### DIFF
--- a/app/models/answer.rb
+++ b/app/models/answer.rb
@@ -5,7 +5,7 @@ class Answer < ApplicationRecord
     CONTEXT_LENGTH_EXCEEDED_RESPONSE = "Sorry, your last question was too complex for me to answer. " \
       "Could you make your question more specific? You can also try splitting it into multiple " \
       "smaller questions and asking them separately.".freeze
-    OPENAI_CLIENT_ERROR_RESPONSE = <<~MESSAGE.freeze
+    ANSWER_SERVICE_ERROR_RESPONSE = <<~MESSAGE.freeze
       Sorry, something went wrong while trying to answer your question. Try again later.
 
       We saved your conversation. Check [GOV.UK guidance for businesses](https://www.gov.uk/browse/business) if you need information now.
--- a/app/models/question.rb
+++ b/app/models/question.rb
@@ -3,6 +3,7 @@ class Question < ApplicationRecord
        {
          open_ai_rag_completion: "open_ai_rag_completion", # legacy strategy - no longer used
          openai_structured_answer: "openai_structured_answer",
+         claude_structured_answer: "claude_structured_answer",
        },
        prefix: true
 
--- a/config/application.rb
+++ b/config/application.rb
@@ -101,5 +101,7 @@ module GovukChat
     config.conversation_js_progressive_disclosure_delay = nil
 
     config.bigquery_dataset_id = ENV["BIGQUERY_DATASET"]
+
+    config.answer_strategy = ENV.fetch("ANSWER_STRATEGY", "openai_structured_answer")
   end
 end
--- a/lib/answer_composition/pipeline/claude/structured_answer_composer.rb
+++ b/lib/answer_composition/pipeline/claude/structured_answer_composer.rb
@@ -1,0 +1,99 @@
+module AnswerComposition::Pipeline::Claude
+  class StructuredAnswerComposer
+    BEDROCK_MODEL = "eu.anthropic.claude-3-5-sonnet-20240620-v1:0".freeze
+
+    def self.call(...) = new(...).call
+
+    def initialize(context)
+      @context = context
+    end
+
+    def call
+      start_time = Clock.monotonic_time
+
+      response = bedrock_client.converse(
+        system: [{ text: system_prompt }],
+        model_id: BEDROCK_MODEL,
+        messages:,
+        inference_config:,
+        tool_config:,
+      )
+
+      context.answer.assign_llm_response("structured_answer", response.to_h)
+      message = response.dig("output", "message", "content", 0, "tool_use", "input", "answer")
+      context.answer.assign_attributes(message:, status: "answered")
+      context.answer.assign_metrics("structured_answer", build_metrics(start_time, response))
+    end
+
+  private
+
+    attr_reader :context
+
+    def messages
+      [
+        {
+          role: "user",
+          content: [{ text: context.question_message }],
+        },
+      ]
+    end
+
+    def inference_config
+      {
+        max_tokens: 1000,
+        temperature: 0.0,
+      }
+    end
+
+    def system_prompt
+      <<~PROMPT
+        You are a chat assistant for the UK government, designed to provide helpful and contextually relevant responses to user queries.
+        Provide concise responses based on the content on the GOV.UK website.
+      PROMPT
+    end
+
+    def bedrock_client
+      @bedrock_client ||= Aws::BedrockRuntime::Client.new
+    end
+
+    def build_metrics(start_time, response)
+      {
+        duration: Clock.monotonic_time - start_time,
+        llm_prompt_tokens: response.dig("usage", "input_tokens"),
+        llm_completion_tokens: response.dig("usage", "output_tokens"),
+      }
+    end
+
+    def tool_config
+      {
+        tools: tools,
+        tool_choice: {
+          tool: {
+            name: "answer_confidence",
+          },
+        },
+      }
+    end
+
+    def tools
+      [
+        {
+          tool_spec: {
+            name: "answer_confidence",
+            description: "Prints the answer of a given question with a confidence score.",
+            input_schema: {
+              json: {
+                type: "object",
+                properties: {
+                  answer: { description: "Your answer to the question in markdown format", title: "Answer", type: "string" },
+                  confidence: { description: "Your confidence in the answer provided, ranging from 0.0 to 1.0", title: "Confidence", type: "number" },
+                },
+                required: %w[answer confidence],
+              },
+            },
+          },
+        },
+      ]
+    end
+  end
+end
--- a/lib/answer_composition/pipeline_runner.rb
+++ b/lib/answer_composition/pipeline_runner.rb
@@ -1,5 +1,5 @@
 module AnswerComposition
-  class OpenAIAnswer
+  class PipelineRunner
     def self.call(...) = new(...).call
 
     def initialize(question:, pipeline: [])
@@ -26,9 +26,16 @@ module AnswerComposition
     rescue OpenAIClient::RequestError => e
       GovukError.notify(e)
       context.abort_pipeline(
-        message: Answer::CannedResponses::OPENAI_CLIENT_ERROR_RESPONSE,
+        message: Answer::CannedResponses::ANSWER_SERVICE_ERROR_RESPONSE,
         status: "error_answer_service_error",
         error_message: error_message(e),
+      )
+    rescue Aws::Errors::ServiceError => e
+      GovukError.notify(e)
+      context.abort_pipeline(
+        message: Answer::CannedResponses::ANSWER_SERVICE_ERROR_RESPONSE,
+        status: "error_answer_service_error",
+        error_message: e.message,
       )
     end
 
--- a/spec/lib/answer_composition/pipeline_runner_spec.rb
+++ b/spec/lib/answer_composition/pipeline_runner_spec.rb
@@ -1,4 +1,4 @@
-RSpec.describe AnswerComposition::OpenAIAnswer do # rubocop:disable RSpec/SpecFilePathFormat
+RSpec.describe AnswerComposition::PipelineRunner do
   describe "#call" do
     let(:question) { build(:question) }
 
@@ -91,7 +91,7 @@ RSpec.describe AnswerComposition::OpenAIAnswer do # rubocop:disable RSpec/SpecFi
           .and have_attributes(
             question:,
             status: "error_answer_service_error",
-            message: Answer::CannedResponses::OPENAI_CLIENT_ERROR_RESPONSE,
+            message: Answer::CannedResponses::OPENAI_ERROR_RESPONSE,
             error_message: "class: OpenAIClient::RequestError message: nested error message",
           )
       end
--- a/spec/lib/answer_composition/pipeline_runner_spec.rb
+++ b/spec/lib/answer_composition/pipeline_runner_spec.rb
@@ -91,7 +91,7 @@ RSpec.describe AnswerComposition::PipelineRunner do
           .and have_attributes(
             question:,
             status: "error_answer_service_error",
-            message: Answer::CannedResponses::OPENAI_ERROR_RESPONSE,
+            message: Answer::CannedResponses::ANSWER_SERVICE_ERROR_RESPONSE,
             error_message: "class: OpenAIClient::RequestError message: nested error message",
           )
       end
@@ -117,6 +117,31 @@ RSpec.describe AnswerComposition::PipelineRunner do
           result = described_class.call(question:, pipeline: [pipeline_step])
           expect(result.error_message).to eq "class: OpenAIClient::RequestError message: default error message"
         end
+      end
+    end
+
+    context "when the step raises an Aws::Errors::ServiceError" do
+      let(:pipeline_step) do
+        client = stub_bedrock_converse("ServerError")
+        ->(_context) { client.converse(model_id: "just-generating-an-error") }
+      end
+
+      it "notifies sentry" do
+        expect(GovukError).to receive(:notify).with(kind_of(Aws::Errors::ServiceError))
+        described_class.call(question:, pipeline: [pipeline_step])
+      end
+
+      it "returns the context's answer with the correct message, status and error_message" do
+        result = described_class.call(question:, pipeline: [pipeline_step])
+
+        expect(result)
+          .to be_a(Answer)
+          .and have_attributes(
+            question:,
+            status: "error_answer_service_error",
+            message: Answer::CannedResponses::ANSWER_SERVICE_ERROR_RESPONSE,
+            error_message: "stubbed-response-error-message",
+          )
       end
     end
   end
--- a/spec/support/stub_bedrock.rb
+++ b/spec/support/stub_bedrock.rb
@@ -1,0 +1,77 @@
+module StubBedrock
+  # Usage scenarios
+  #
+  # Stub question message provides specific answer
+  #     client = stub_bedrock_converse(
+  #       bedrock_claude_structured_answer_response("Expected question", "Expected answer")
+  #     )
+  #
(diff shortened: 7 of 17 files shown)


config/application.rb
```diff
@@ -101,5 +101,7 @@ class Application < Rails::Application
     config.conversation_js_progressive_disclosure_delay = nil
 
     config.bigquery_dataset_id = ENV["BIGQUERY_DATASET"]
+
+    config.answer_strategy = ENV["ANSWER_STRATEGY"]
```
kevindew, line 105:
```suggestion
    config.answer_strategy = ENV.fetch("ANSWER_STRATEGY", "openai_structured_answer")
```

so there is a default set automatically.

spec/support/stub_bedrock_request.rb
```diff
@@ -0,0 +1,27 @@
+module StubBedrockRequest
```
kevindew, line 1: So I've had a bit of a think about what things we may want from stubbing and done some prototyping to build on your work.

My thinking is that we want the stubbing so:

1. (most basic) we can write tests that pass
2. want to cut out verbosity of our individual tests
3. want to verify a basic flow of providing minimal input and output for an assertion.

This is the rough code I've put together (note comments are more for your individual benefit than for prosperity):

```ruby
module StubBedrock
  # Usage scenarios
  # 
  # ## Stub question message provides specific answer
  #
  # client = stub_bedrock_converse(
  #   bedrock_structured_answer_response("Expected question", "Expected answer")
  # )
  #
  # ## Simulate basic text response
  # client = stub_bedrock_converse(
  #   bedrock_text_response("VAT (Value Added Tax) is a tax applied to most goods and services in the UK.")
  # )
  #
  # ## 
  #
  # ## Simulate an error
  # ```
  # client = stub_bedrock_converse("NotFound")
  # client.converse(model_id: "just-generating-an-error")
  # => Aws::BedrockRuntime::Errors::ServerError: stubbed-response-error-message
  # ```
  def stub_bedrock_converse(*responses)
    bedrock_client = Aws::BedrockRuntime::Client.new(stub_responses: true)
    allow(Aws::BedrockRuntime::Client).to receive(:new).and_return(bedrock_client)
    bedrock_client.stub_responses(:converse, responses)
    bedrock_client
  end

  # Intended as a high level mock that will provide a specific response only
  # if given a specific question - I imagine more specialised usages (e.g.
  # asserting function calls) would be definied in a test file
  def bedrock_structured_answer_response(question, answer)
    lambda do |context|
      given_question = context.params.dig(:messages, 0, :content, 0, :text)

      if question && given_question != question
        raise "Unexpected question received: \"#{given_question}\". Expected \"#{question}\"."
      end
      
      bedrock_text_response(question)
    end
  end

  # Intended as a lower level mocked response that could be used in unit tests
  # to consider more data that is returned in the response than just the answer content
  def bedrock_text_response(text, input_tokens: 10, output_tokens: 20, latency_ms: 999)
    {
      output: {
        message: {
          role: "assistant",
          content: [{ text: }],
        },
      },
      stop_reason: "end_turn",
      usage: {
        input_tokens:,
        output_tokens:,
        total_tokens: input_tokens + output_tokens,
      },
      metrics: {
        latency_ms:,
      },
    }
  end
end
```

spec/lib/answer_composition/pipeline_runner_spec.rb
```diff
@@ -119,5 +119,33 @@
...
     end
+
+    context "when the step raises an Aws::BedrockRuntime::Errors::ValidationException" do
+      let(:error) do
+        Aws::BedrockRuntime::Errors::ValidationException.new(
+          Seahorse::Client::RequestContext.new,
+          "The model returned the following errors: Input is too long for requested model.",
+        )
+      end
+      let(:pipeline_step) { ->(_context) { raise error } }
```
kevindew, line 130: As this means reaching a bit into how AWS exceptions work I think it might be better to use the stubbed client. E.g:

```
let(:pipeline_step) do
  client = stub_bedrock_converse("ServerError")
  ->(_context) { client.converse(model_id: "just-generating-an-error") }
end
```

It feels a bit more of a true representation of what we'd get by using the client - so long as we can do it with low verbosity.

spec/lib/answer_composition/pipeline/bedrock_structured_answer_composer_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+      ]
+    end
+
+    it "makes a successful request to Bedrock" do
+      request = stub_bedrock_request(:converse, response)
+      described_class.call(context)
+      expect(request[0][:http].status_code).to be 200
+    end
+
+    context "when a successful response is received" do
```
kevindew, line 35: I don't think you need this context as our only context is for a successful response - we don't have code for it being unsuccessful.

lib/answer_composition/pipeline/bedrock_structured_answer_composer.rb
```diff
@@ -0,0 +1,99 @@
...
+      response = bedrock_client.converse(
+        system: [{ text: system_prompt }],
+        model_id: BEDROCK_MODEL,
+        messages:,
+        inference_config:,
+        tool_config:,
+      )
+
+      context.answer.assign_llm_response("structured_answer", response.to_h)
+      message = response["output"]["message"]["content"][0]["tool_use"]["input"]["answer"]
```
kevindew, line 23:
```suggestion
      message = response.dig("output", "message", "content", 0, "tool_use", "input", "answer")
```

using dig should be a bit cleaner. 

So we don't need to do a JSON parse, it merges the JSON into the response rather than returning a string of JSON?

spec/support/stub_bedrock_request.rb
kevindew: Note: I wrote the above before you'd pushed the JSON up

spec/lib/answer_composition/pipeline/bedrock_structured_answer_composer_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+      it "calls Bedrock converse endpoint and assigns the correct values to the context's answer" do
```
kevindew, line 36:
```suggestion
      it "uses Bedrock converse endpoint to assign the correct values to the context's answer" do
```

Just a minor edit to reflect that Bedrock is used to populate the answer.

---

This test seems good. I think a decision for us is whether we want to just assert that given a particular question we get an answer back (as per the method I suggested for the test helper) or whether we want to assert any of the data that is pushed to bedrock (tool configuration for example)

I think it's ok that we don't assert the tool config - as it is config after all - but do you know if that being invalid would raise an exception? I'm kinda hoping there isn't a need to assert it because AWS validates it has the right object structure.

spec/lib/answer_composition/pipeline/bedrock_structured_answer_composer_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+        stub_bedrock_request(:converse, response)
+
+        described_class.call(context)
+
+        expect(context.answer.metrics["structured_answer"]).to eq({
+          duration: 1.5,
+          llm_prompt_tokens: 10,
+          llm_completion_tokens: 20,
+        })
+      end
```
kevindew, line 68: This test looks good too - hopefully we can use a stub for this rather than needing to define the full response so it's a bit clearer where the token values come from.

spec/lib/answer_composition/pipeline/bedrock_structured_answer_composer_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+    end
+  end
+
```
kevindew, line 71: I'm guessing the bits below here are intended to be deleted in favour of using shared helpers?

kevindew (commented):
Looks great Chae, sorry for taking a while to get this review together.

I think the other thing to add which we discussed was a system test similar to https://github.com/alphagov/govuk-chat/blob/main/spec/system/conversation_with_open_ai_with_structured_answer_spec.rb

spec/lib/answer_composition/pipeline/bedrock_structured_answer_composer_spec.rb
chaecramb: Yeah, I realised I accidentally re-introduced this in a rebase. It's been removed again now, but the commits still need a little cleaning up.

lib/answer_composition/pipeline/bedrock_structured_answer_composer.rb
```diff
@@ -0,0 +1,61 @@
+module AnswerComposition::Pipeline
+  class BedrockStructuredAnswerComposer
```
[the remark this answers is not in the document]
chaecramb: That makes sense.

I was wondering about using Claude vs. Bedrock for naming, since in theory we should be able to use the `converse` endpoint in such a way that it isn't Claude specific and would work with any Bedrock model.

app/models/question.rb
```diff
@@ -3,6 +3,7 @@ class Question < ApplicationRecord
        {
          open_ai_rag_completion: "open_ai_rag_completion", # legacy strategy - no longer used
          openai_structured_answer: "openai_structured_answer",
+         bedrock_structured_answer: "bedrock_structured_answer",
```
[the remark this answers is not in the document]
chaecramb: Yeah, I was wondering about this. I started out using Claude but then switched to Bedrock, since in theory we should be able to use the `converse` endpoint in such a way that it isn't Claude specific and would work with any Bedrock model.

But perhaps the idea that "bedrock_structured_answer" could work with any model simply with an env var change is a bit unrealistic.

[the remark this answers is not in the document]
kevindew: Yeah I think it's very unlikely that the answer strategy alone will suffice with a bedrock name - and would need a claude one.

It does however seem a possibility that for the pipeline classes that they might be fine with a namespace of Bedrock and model specificities could be passed into them at creation (e.g. as below) but I think we might want to start specific and then roll back should a more general case rear it's head - that approach always seems safer than starting generic but not having a second test case and then having something that pretends to be generic but is actually a bit specific.

```
      when "claude_structured_answer"
        PipelineRunner.call(question:, pipeline: [
          Pipeline::Bedrock::StructuredAnswerComposer.new(:claude),
        ])
      when "nova_structured_answer"
        PipelineRunner.call(question:, pipeline: [
          Pipeline::Bedrock::StructuredAnswerComposer.new(:claude),
        ])
```

lib/answer_composition/pipeline/bedrock_structured_answer_composer.rb
chaecramb: Yes, the client returns the response as nested structs.

spec/lib/answer_composition/pipeline/bedrock_structured_answer_composer_spec.rb
chaecramb: Yeah, from everything I've seen so far, AWS always validates the structure of anything pushed to Bedrock, and returns an error if it isn't what's expected/required.

spec/lib/answer_composition/pipeline/bedrock_structured_answer_composer_spec.rb
chaecramb: It seems we do need the full response, the client returns an argument error for missing parameters if you miss out any of the others.

app/models/answer.rb
```diff
@@ -5,7 +5,12 @@ module CannedResponses
     CONTEXT_LENGTH_EXCEEDED_RESPONSE = "Sorry, your last question was too complex for me to answer. " \
       "Could you make your question more specific? You can also try splitting it into multiple " \
       "smaller questions and asking them separately.".freeze
-    OPENAI_CLIENT_ERROR_RESPONSE = <<~MESSAGE.freeze
+    OPENAI_ERROR_RESPONSE = <<~MESSAGE.freeze
```
kevindew, line 8: I imagine we don't need this env var and can just ANSWER_SERVICE_ERROR_RESPONSE when this is used?

chaecramb: Yeah, that was my intention with renaming previously and using the same error message for both Claude and OpenAI. Will remove it again and use the `ANSWER_SERVICE_ERROR_RESPONSE` variable name.

lib/answer_composition/pipeline_runner.rb
```diff
@@ -26,10 +26,17 @@ def call
     rescue OpenAIClient::RequestError => e
       GovukError.notify(e)
       context.abort_pipeline(
-        message: Answer::CannedResponses::OPENAI_CLIENT_ERROR_RESPONSE,
+        message: Answer::CannedResponses::ANSWER_SERVICE_ERROR_RESPONSE,
         status: "error_answer_service_error",
         error_message: error_message(e),
       )
+    rescue Aws::Errors::ServiceError => e
```
jackbot, line 33: Given this is now a generic `PipelineRunner` class - rather than being an OpenAI specific class - would it be favourable to throw generic errors from the implementation classes themselves and catch those, rather than catching implementation-specific errors here?

So in the `AnswerComposition::Pipeline::Claude` we'd catch the `Aws::Errors::ServiceError` and throw an `AnswerComposition::ServiceError` (or something) error.

Same thing with the OpenAI stuff; we'd catch `OpenAIClient::ContextLengthExceededError` in `AnswerComposition::Pipeline::OpenAIStructuredAnswerComposer` and throw `AnswerComposition::ContentLengthExceededError`.

lib/answer_composition/pipeline/claude/structured_answer_composer.rb
```diff
@@ -0,0 +1,99 @@
...
+    end
+
+    def inference_config
+      {
+        max_tokens: 1000,
+        temperature: 1.0,
+      }
+    end
+
+    def system_prompt
```
jackbot, line 48: Am I right in assuming we're hard-coding these prompts and tool call configs temporarily while we test the implementation, and that we'd want to put these in `govuk_chat_private` eventually?

spec/support/stub_bedrock.rb
```diff
@@ -0,0 +1,76 @@
...
+  # client = stub_bedrock_converse(
+  #   bedrock_claude_structured_answer_response("Expected question", "Expected answer")
+  # )
+  #
+  # ## Simulate basic tool response
+  # client = stub_bedrock_converse(
+  #   bedrock_claude_tool_response({ "key" => "value" }, tool_user
+  # )
+  #
+  # ##
```
jackbot, line 15: Looks like an extra blank heading here.

spec/support/stub_bedrock.rb
```diff
@@ -0,0 +1,76 @@
...
+  #
+  # ## Simulate an error
+  # ```
```
jackbot, line 18: A bit nit-picky but the two code examples above aren't surrounded by the three backticks. Not sure they're fully necessary as it's pretty clear the examples are all code.

spec/support/stub_bedrock.rb
```diff
@@ -0,0 +1,76 @@
...
+  # client = stub_bedrock_converse("NotFound")
+  # client.converse(model_id: "just-generating-an-error")
+  # => Aws::BedrockRuntime::Errors::ServerError: stubbed-response-error-message
+  # ```
+  def stub_bedrock_converse(*responses)
```
jackbot, line 23: Do you think it's worth adding an example with multiple responses? It's not fully clear to me how that might work.

lib/answer_composition/pipeline_runner.rb
chaecramb: Yeah, @kevindew and I did talk about doing that but decided on the quickest, easiest implementation to start, with the intention to abstract it later.

Although tbh, I could add it here, as it wouldn't take long.

jackbot: Cool, sounds like a job for later on then :+1:

spec/support/stub_bedrock.rb
chaecramb: @jackbot I've added an example, does this help: https://github.com/alphagov/govuk-chat/blob/62712021b4ca3bd85cd5f1820066291e7aef5ee6/spec/support/stub_bedrock.rb#L11-L24

spec/support/stub_bedrock.rb
```diff
@@ -0,0 +1,92 @@
...
+  #   bedrock_claude_tool_response({ "key" => "value" }, tool_name)
+  # )
+  # ```
+  #
+  # ## Simulate an error
+  # ```
+  # client = stub_bedrock_converse("NotFound")
+  # client.converse(model_id: "just-generating-an-error")
+  # => Aws::BedrockRuntime::Errors::ServerError: stubbed-response-error-message
+  # ```
```
kevindew, line 38:
```suggestion
  # Usage scenarios
  #
  # Stub question message provides specific answer
  #     client = stub_bedrock_converse(
  #       bedrock_claude_structured_answer_response("Expected question", "Expected answer")
  #     )
  #
  # Stub multiple responses
  #     client = stub_bedrock_converse(
  #       bedrock_claude_structured_answer_response("Expected question 1", "Expected answer 1")
  #       bedrock_claude_structured_answer_response("Expected question 2", "Expected answer 2")
  #     )
  #
  # Simulate basic tool response
  #     client = stub_bedrock_converse(
  #       bedrock_claude_tool_response({ "key" => "value" }, tool_name)
  #     )
  #
  # Simulate an error
  #     client = stub_bedrock_converse("NotFound")
  #     client.converse(model_id: "just-generating-an-error")
  #     => Aws::BedrockRuntime::Errors::ServerError: stubbed-response-error-message
```

kevindew: Suggestion above is to clean up the formatting a bit since I probably made things a bit weird with the markdown. I think indentation should suffice to separate code out.

kevindew (commented):
Looks great, just made a suggestion to the stub_bedrock comment to clear it up then good to merge.

I've tried it out locally and it works great.

spec/support/stub_bedrock.rb
chaecramb: I've switched it to this

kevindew (approved):
Great, let's go 🚀

PR #31, opened 2025-02-03
Export and configure AWS credentials

jackbot:
In order to start using Bedrock by making direct API calls to it, we
need to authenticate those requests with some credentials.

We can use dotenv here to load a specific file - in this case
`.env.aws.local` which will contain all the necessary env vars to allow
the Bedrock gem to authenticate.

There's a new script in the `bin` directory which uses gds-cli to
generate the credentials and write them to this file. A bit of filtering
on this output needs to be done though, as the CLI outputs a blob of
text like this:

```
unset AWS_CREDENTIALS_FILE;
unset AWS_DEFAULT_PROFILE;
unset AWS_PROFILE;
export AWS_REGION='eu-west-1';
...
```

Dotenv doesn't like the `unset` lines, and it includes the trailing
semicolon as part of the value in the `export` lines which makes the
credentials invalid.

So the `export-aws-credentials` script filters for just the lines
beginning with `export` and strips the trailing semicolon before writing
to the file.

The `bin/setup` script has been extended so it calls the above export
script. This just helps set everything up for developers using that
setup script.

Finally a new `bin/launch-in-govuk-docker` script is added for
convenience. It calls the export script then runs `govuk-docker up` with
the `app` stack as default and passes all the script arguments to it.


bin/export-aws-credentials
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env ruby
+
+puts "== Configuring AWS credentials =="
+
+rails_root = File.expand_path("..", __dir__)
+config_file_path = File.join(rails_root, ".env.aws.local")
+
+# Export AWS credentials using the gds-cli tool, stripping off the leading "export" and the trailing ";"
+system "gds aws govuk-test-poweruser -e --art 12h | grep '^export' | sed -E 's/^export //; s/;$//' > #{config_file_path}"
```
kevindew, line 9: I get an error on this because 12h exceeds the maximum time. I can get up to 9h, but suggest we set this at 8h since that should cover a working day.

bin/export-aws-credentials
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env ruby
+
+puts "== Configuring AWS credentials =="
+
+rails_root = File.expand_path("..", __dir__)
+config_file_path = File.join(rails_root, ".env.aws.local")
```
kevindew, line 6: Nice that this path is already gitignored, did that influence your choice of a `.local` suffix? As I was going to ask about that

config/application.rb
```diff
@@ -18,6 +18,8 @@
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
 
+Dotenv::Rails.files << ".env.aws.local" if ENV["RAILS_ENV"] == "development"
```
kevindew, line 21: I'd have thought we could do `Rails.env.development?` - however I do wonder if we think there are any risks in having this without the conditional? As we don't have the same conditional for dotenv itself which could also load env vars

bin/launch-in-govuk-docker
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env ruby
```
kevindew, line 1: I think rather than having this as an individual file we should see if we can combine it with setup. I'll just have a quick look at how I think that could work.

kevindew (commented):
Nice job - I think there's a few things to sort out but this gets us most of the way there. I'll just have a little think about whether a unified bin/setup could work.

bin/export-aws-credentials
jackbot: That's very strange, the 12 hour timeframe works fine for me...do you know why that might be?

I'll change to 8h in any case.

bin/export-aws-credentials
jackbot: I thought a `.local` suffix made sense to make it obvious that it's a file for a dev setup rather than any other env. It was a nice coincidence that it was already gitignored.

config/application.rb
kevindew: Turns out we do need the conditional (sorry) because dotenv is only in our development and test Rails groups so we get an exception for a production gem install:

```
 !     NameError: uninitialized constant Dotenv (NameError)
 !     
 !     Dotenv::Rails.files << ".env.aws.local"
 !     ^^^^^^
 !     /tmp/build_777281f8/config/application.rb:21:in '<main>'
 !     /tmp/build_777281f8/Rakefile:6:in 'Kernel#require_relative'
 !     /tmp/build_777281f8/Rakefile:6:in '<main>'
```

(sourced from Heroku)

bin/export-aws-credentials
kevindew: How interesting no I don't. I wonder if we increased the time policy at some point and it only applies to new accounts.

config/application.rb
jackbot: No worries. Added it back in with the more Rails-y version.

bin/export-aws-credentials
kevindew: One thing I've noticed is that if you already have credentials and run it with longer time frames it seems to just spit out the same ones again, presumably ignoring the long time window.

bin/launch-in-govuk-docker
kevindew: So I had a quick play around and got to this:

```
#!/usr/bin/env ruby
require "fileutils"

APP_ROOT = File.expand_path("..", __dir__)

def system!(*args)
  system(*args, exception: true)
end

FileUtils.chdir APP_ROOT do
  # This script is a way to set up or update your development environment automatically.
  # This script is idempotent, so that you can run it at any time and get an expectable outcome.

  # only run this on local machine, never on govuk-docker
  system! "bin/export-aws-credentials"

  govuk_docker = ARGV.include?("--govuk-docker") || ARGV.include?("-d")

  command_prefix = govuk_docker ? "govuk-docker run govuk-chat-lite " : ""

  puts "== Installing dependencies =="
  system("#{command_prefix}bundle check") || system!("#{command_prefix}bundle install")

  puts "\n== Preparing database =="
  system! "#{command_prefix}bin/rails db:prepare"

  puts "\n== Removing old logs and tempfiles =="
  system! "#{command_prefix}bin/rails log:clear tmp:clear"

  unless ARGV.include?("--skip-server")
    puts "\n== Starting development server =="
    STDOUT.flush # flush the output before exec(2) so that it displays

    # expect that devs don't intend to start queue_consumer when running the app
      # under normal circumstances
    if govuk_docker
      # avoid bin/dev in docker to avoid a single docker process managing multiple processes
      exec "govuk-docker up govuk-chat-app govuk-chat-css govuk-chat-worker"
    else
      exec "bin/dev -m all=1,queue_consumer=0"
    end
  end
end

```

I figured it was best to move the aws credentials as the first task because that requires interactivity (MFA) whereas the rest doesn't.

jackbot: Nice one.

The comment on line 14 says "only run this on local machine, never on govuk-docker" but there's no conditional on the system command itself. 

I guess we don't want to run that in Docker because it requires interactivity, so should we move it below the `govuk_docker` variable and only run it if that's `true`?

bin/export-aws-credentials
kevindew: Sorry I forgot to click unresolve on this on Monday so I think you'll have missed this comment.

kevindew (commented):
Sorry Jack for leaving you hanging for the comments here.

bin/launch-in-govuk-docker
jackbot: Got it, that makes sense :+1:

bin/export-aws-credentials
jackbot: Cheers, i've updated the script now so it'll raise an error on failure.

bin/setup
```diff
@@ -10,25 +10,34 @@ end
...
+
+  govuk_docker = ARGV.include?("--govuk-docker") || ARGV.include?("-d")
 
-  # puts "\n== Copying sample files =="
-  # unless File.exist?("config/database.yml")
-  #   FileUtils.cp "config/database.yml.sample", "config/database.yml"
-  # end
+  command_prefix = govuk_docker ? "govuk-docker run govuk-chat-lite " : ""
+
+  puts "== Installing dependencies =="
```
kevindew, line 21:
```suggestion
  puts "\n== Installing dependencies =="
```
Super minor nit pick for a bit of extra spacing.

kevindew (commented):
Great stuff, I tried it out locally and works really well. I'm quite excited about trying to adopt `bin/setup`

Just a couple of comments to resolve.

README.md
```diff
@@ -31,6 +31,16 @@ Copy the `.env.example` file to `.env` and ask a team member for the values to u
 cp .env.example .env
 ```
 
+### Authenticating with AWS
```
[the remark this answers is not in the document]
jackbot: Cheers for that, I've updated with your suggestions.

README.md
```diff
@@ -45,16 +43,36 @@ bundle config --local github.com <token>
...
 
+Or can be run in govuk-docker with:
+
+```bash
+govuk-docker run govuk-chat-lite bundle exec rake
+```
+
+### Starting the app
+
+To start app in govuk-docker from your local machine:
```
kevindew, line 64:
```suggestion
To start the app in govuk-docker from your local machine:
```

kevindew (approved):
Super, just a couple of minor things that shouldn't need another review

PR #34, opened 2025-02-04
Extend LinkTokenMapper to handle anchor tags

jackbot:
https://trello.com/c/y327T4Dy/2270

We had an issue recently where the `LinkTokenMapper` returned an invalid
URL in an answer because the source URL was an anchor tag.

For context, the `LinkTokenMapper` will read some HTML and store the links
in a hash with a key of the URL and value of "link_x" where x is an
integer.

A method can then be called to pass in some markdown which contains
links in the form "link_x" and it rewrites those to the stored URLs.

However there's a bug when an anchor link is used. In this case, the
mapper will take a link like "#heading-1" and rewrite it to
"https://www.gov.uk/#heading-1".

This happens because we store the link's `href` in the mapping, then
when we come to replace the token with a link, it's seen as a relative
URL so we turn it into an absolute URL by prepending the website root to
it.

This works fine for relative URLs but not for anchor tags. In those
cases, we need to know the `exact_path` of the chunk we're using so we
can make it into a relative URL.

For example, if we have a chunk with the `exact_path` of
"/browse/business" and some HTML content of:

```
this is <a href="#anchor">an anchor</a>.
```

then when we come to replace the token with a link it needs to produce
the URL `https://www.gov.uk/browse/business#anchor`, rather than
`https://www.gov.uk/#anchor` which it does now.

To fix this we can pass the `exact_path` as a parameter in when mapping
the links to tokens. If the link is an anchor (i.e. starts with a "#")
then we store the exact path plus the anchor in the mapping.

The rest of the code doesn't need to change because when we replace the
tokens with links, all the stored links are either full relative links
or absolute links, there won't be any plain anchor tags.

### DIFF
--- a/lib/answer_composition/link_token_mapper.rb
+++ b/lib/answer_composition/link_token_mapper.rb
@@ -6,11 +6,15 @@ module AnswerComposition
       @mapping = {}
     end
 
-    def map_links_to_tokens(html_content)
+    def map_links_to_tokens(html_content, exact_path)
       doc = Nokogiri::HTML::DocumentFragment.parse(html_content)
 
       doc.css("a").each do |link|
-        href = link["href"]
+        href = begin
+          URI.join(ensure_absolute_govuk_url(exact_path), link["href"]).to_s
+        rescue URI::InvalidURIError
+          link["href"]
+        end
 
         if mapping.key?(href)
           link["href"] = mapping[href]
@@ -61,9 +65,7 @@ module AnswerComposition
 
       if (url = link_for_token(token))
         link_element.tap do |el|
-          # We frequently host GOV.UK chat in environments off www.gov.uk
-          # and need links not to be relative so that they will work.
-          el.attr["href"] = ensure_absolute_govuk_url(url)
+          el.attr["href"] = url
 
           # If we have a link where the text is e.g. "link_1" then we should replace
           # it with "source". Showing "link_1" to the user makes it seem like something
@@ -82,11 +84,9 @@ module AnswerComposition
     end
 
     def ensure_absolute_govuk_url(url)
-      begin
-        relative_uri = URI(url)
-      rescue URI::InvalidURIError
-        return url
-      end
+      # We frequently host GOV.UK chat in environments off www.gov.uk
+      # and need links not to be relative so that they will work.
+      relative_uri = URI(url)
 
       return url if relative_uri.absolute?
 
--- a/lib/answer_composition/pipeline/openai_structured_answer_composer.rb
+++ b/lib/answer_composition/pipeline/openai_structured_answer_composer.rb
@@ -89,7 +89,7 @@ module AnswerComposition::Pipeline
           page_title: result.title,
           page_description: result.description,
           context_headings: result.heading_hierarchy,
-          context_content: link_token_mapper.map_links_to_tokens(result.html_content),
+          context_content: link_token_mapper.map_links_to_tokens(result.html_content, result.exact_path),
         }
       end
     end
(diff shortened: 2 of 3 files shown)


lib/answer_composition/link_token_mapper.rb
```diff
@@ -6,11 +6,14 @@ def initialize
...
 
-    def map_links_to_tokens(html_content)
+    def map_links_to_tokens(html_content, exact_path)
       doc = Nokogiri::HTML::DocumentFragment.parse(html_content)
 
       doc.css("a").each do |link|
         href = link["href"]
+        if href.start_with?("#")
+          href = exact_path + href
+        end
```
kevindew, line 16: So a consideration we should make here is that the exact path could already have a fragment and we have to replace it.

kevindew: I imagine we want to see if we can use a URI function for this then we can be super generic.

From a quick poke there is `URI::join` but that requires an absolute URI as the first input but aside from that does all the things we'd want pretty neatly:

```
[7] pry(main)> URI::join("https://www.gov.uk/already-got-a-fragment#fragment", "#new-fragment")
=> #<URI::HTTPS https://www.gov.uk/already-got-a-fragment#new-fragment>
[8] pry(main)> URI::join("https://www.gov.uk/no-fragment", "#appends-fragment")
=> #<URI::HTTPS https://www.gov.uk/no-fragment#appends-fragment>
[9] pry(main)> URI::join("https://www.gov.uk/replaces/path", "/new-absolute-path")
=> #<URI::HTTPS https://www.gov.uk/new-absolute-path>
[10] pry(main)> URI::join("https://www.gov.uk/bonus-for/relative-path#with-a-fragment", "new-relative-path")
=> #<URI::HTTPS https://www.gov.uk/bonus-for/new-relative-path>
```

jackbot: Cheers. I've made a change to use `URI.join`.

kevindew: It'd be good if this didn't need the fragment conditional so all URIs are treated the same as absolute.

There's also a bit of an edge case where `ensure_absolute_govuk_url` can return an invalid URI that URI.join would choke on.

jackbot: The only way I can see to achieve this without a conditional is like this:

```ruby
doc.css("a").each do |link|
  absolute_url = ensure_absolute_govuk_url(link["href"])
  href = begin
    URI.join(absolute_url, exact_path, link["href"]).to_s
  rescue URI::InvalidURIError
    absolute_url
  end

  if mapping.key?(href)
    link["href"] = mapping[href]
  else
    token = map_link_to_token(href)
    link["href"] = token
  end
end
```

So we first "absolutise" the URL then we call `URI.join` with the absolute URL, `exact_path` and `link["href"]`. This covers the three cases that we need to handle:

```
with exact_path of "/exact-path"

https://www.gov.uk/search  ->  https://www.gov.uk/search
/browse/business -> https://www.gov.uk/browse/business
#heading -> https://www.gov.uk/exact-path#heading 
```

The issue being that we only want to use the exact path if the `link['href']` is an anchor tag, otherwise we want to ignore it. So the code above essentially appends the `exact_path`, then appends `link['href']`, which will overwrite the path if `link['href']` is anything other than an anchor.

It works, but it feels like the code is very confusing. Is this the kind of thing you had in mind?

kevindew: Thanks, good points - I've just been having a ponder. Though may well be better to have a chat. I'll put some quick ideas here and reach out.

So I imagine that we can do something simpler than what you said but I could be wrong. Something like:

```
      doc.css("a").each do |link|
        begin
          # link uri relative to exact path
          href = URI.join(ensure_absolute_govuk_url(exact_path), link["href"]).to_s
        rescue rescue URI::InvalidURIError
          href = link["href"]
        end
```

and we could remove the URI::InvalidError handling from `ensure_absolute_govuk_url`. 

As far as I can tell this would cover all scenarios?

PR #40, opened 2025-02-07
Implement basic RAG with Claude

chaecramb:
This PR iterates on the Claude answer strategy to add fetching search results to the Pipeline and then replicating the structured answer approach used for OpenAI.

The Claude `structured_answer_composer` now provides the retrieved sources in the request to Claude, incorporates links in the answer, and marks sources as used or unused.

### DIFF
--- a/lib/answer_composition/pipeline/claude/structured_answer_composer.rb
+++ b/lib/answer_composition/pipeline/claude/structured_answer_composer.rb
@@ -51,19 +51,10 @@ module AnswerComposition::Pipeline
       end
 
       def system_prompt
-        sprintf(prompt_config[:system_prompt], context: system_prompt_context)
-      end
-
-      def system_prompt_context
-        context.search_results.map do |result|
-          {
-            page_url: link_token_mapper.map_link_to_token(result.exact_path),
-            page_title: result.title,
-            page_description: result.description,
-            context_headings: result.heading_hierarchy,
-            context_content: link_token_mapper.map_links_to_tokens(result.html_content, result.exact_path),
-          }
-        end
+        sprintf(
+          prompt_config[:system_prompt],
+          context: context.search_results_prompt_formatted(link_token_mapper),
+        )
       end
 
       def prompt_config
--- a/lib/answer_composition/pipeline/claude/structured_answer_composer.rb
+++ b/lib/answer_composition/pipeline/claude/structured_answer_composer.rb
@@ -1,99 +1,109 @@
-module AnswerComposition::Pipeline::Claude
-  class StructuredAnswerComposer
-    BEDROCK_MODEL = "eu.anthropic.claude-3-5-sonnet-20240620-v1:0".freeze
+module AnswerComposition::Pipeline
+  module Claude
+    class StructuredAnswerComposer
+      BEDROCK_MODEL = "eu.anthropic.claude-3-5-sonnet-20240620-v1:0".freeze
 
-    def self.call(...) = new(...).call
+      def self.call(...) = new(...).call
 
-    def initialize(context)
-      @context = context
-    end
+      def initialize(context)
+        @context = context
+        @link_token_mapper = AnswerComposition::LinkTokenMapper.new
+      end
 
-    def call
-      start_time = Clock.monotonic_time
+      def call
+        start_time = Clock.monotonic_time
 
-      response = bedrock_client.converse(
-        system: [{ text: system_prompt }],
-        model_id: BEDROCK_MODEL,
-        messages:,
-        inference_config:,
-        tool_config:,
-      )
+        response = bedrock_client.converse(
+          system: [{ text: system_prompt }],
+          model_id: BEDROCK_MODEL,
+          messages:,
+          inference_config:,
+          tool_config:,
+        )
 
-      context.answer.assign_llm_response("structured_answer", response.to_h)
-      message = response.dig("output", "message", "content", 0, "tool_use", "input", "answer")
-      context.answer.assign_attributes(message:, status: "answered")
-      context.answer.assign_metrics("structured_answer", build_metrics(start_time, response))
-    end
+        tool_output = response.dig("output", "message", "content", 0, "tool_use", "input")
+        set_context_sources(tool_output["sources_used"])
 
-  private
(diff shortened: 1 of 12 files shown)


lib/answer_composition/pipeline/claude/structured_answer_composer.rb
```diff
@@ -69,31 +95,14 @@ def tool_config
         tools: tools,
         tool_choice: {
```
kevindew, line 96: I'm a bit confused about what is supposed to be the structure of this API argument and whether AWS and Claude docs differ or on it's usage (for context I was wondering about whether the tool call is mandatory).

The Claude API docs seem to imply that this tool_choice input should be an object with keys of "name" and "type" as per: https://docs.anthropic.com/en/api/messages#body-tool-choice rather than the tool object provided.

kevindew: I've had my confusion resolved by learning this is part of the Bedrock converse API: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ToolChoice.html

What is quite interesting there, and something that could affect us on question routing, is that they don't seem to have a control for whether parallel tools can be called or not and their text of "any" implies that they'd allow parallel tool calls - which may mean us needing to fall back to invokeModel for question routing. Will add this as questions for AWS day next week.

lib/answer_composition/pipeline/claude/structured_answer_composer.rb
```diff
@@ -40,22 +45,43 @@ def messages
...
+    def system_prompt_context
+      context.search_results.map do |result|
+        {
+          page_url: link_token_mapper.map_link_to_token(result.exact_path),
+          page_title: result.title,
+          page_description: result.description,
+          context_headings: result.heading_hierarchy,
+          context_content: link_token_mapper.map_links_to_tokens(result.html_content, result.exact_path),
+        }
+      end
```
kevindew, line 66: Ah I imagine this is the sort of code where there is abstraction opportunity.

I imagine we could have a really basic class that set-up this context and managed the intersection with link token mapper.

kevindew (commented):
Just a couple of quick comments that don't really affect progress or course

lib/answer_composition/pipeline/openai_structured_answer_composer.rb
kevindew, line 1: If we're editing this file shall we also rename it to be AnswerComposition::Pipeline::OpenAI::StructuredAnswerComposer ? just to start evening up the namespacing.

lib/answer_composition/pipeline/system_prompt_context_builder.rb
```diff
@@ -0,0 +1,26 @@
+module AnswerComposition
+  module Pipeline
```
kevindew, line 2: I think it's a little inconsistent for this class to be in pipeline and it should be in AnswerComposition directly like the link token mapper

lib/answer_composition/pipeline/system_prompt_context_builder.rb
```diff
@@ -0,0 +1,26 @@
...
+    class SystemPromptContextBuilder
```
kevindew, line 3: I think we should have a little bit more of a think about the purpose of this class and it's name. At the moment it feels like we've just moved the code here out the way and I think we can perhaps get its role a little more defined. Will just try have a think over lunch. 

Things I'm wondering are:

- If this is just for search results to be presented in a certain way, should it be just a method on our Context object of: `context.search_results_prompt_formmated(link_token_mapper)`
- should we actually have an object that abstracts away the link_token_mapper as well e.g:

```
manager = StructuredAnswerManager.new(system_prompt_template, search_results)
manager.system_prompt
result = manager.process_response(response)
result.sources_urls
result.message
```

What's your thoughts? Of those two suggestions I think the latter reduces more code at the expense of potentially getting complex if any assumptions fail.

---

oh and a minor aside, we should probably convert the array of search results into a string ourselves whereas I think we're doing it implicitly.

chaecramb: Yeah, I think that makes sense. 

I do quite like the second option but is perhaps excessive for current needs and this pull request.

I've pushed a new commit that moves the search result formatting into `Context`.

lib/answer_composition/pipeline/claude/structured_answer_composer.rb
```diff
@@ -69,31 +85,14 @@ def tool_config
         tools: tools,
         tool_choice: {
           tool: {
-            name: "answer_confidence",
+            name: "output_schema",
```
kevindew, line 88: For a follow up PR: can we rename this to be a bit more descriptive of the tool. I'd suggest "structured_answer". I can see how we got the name from the OpenAI route but it feels like a confusing name for a tool as it is.

lib/answer_composition/pipeline/context.rb
```diff
@@ -68,5 +68,20 @@ def update_sources_from_exact_paths_used(exact_paths)
 
       answer.sources = used_sources + unused_sources
     end
+
+    def search_results_prompt_formatted(link_token_mapper)
```
kevindew, line 72: This looks good, we should add a unit test for it.

kevindew (commented):
Great - I think just some tests to add and then a follow up task if you agree.

lib/answer_composition/pipeline/claude/structured_answer_composer.rb
```diff
@@ -40,22 +44,34 @@ def messages
...
-        Provide concise responses based on the content on the GOV.UK website.
-      PROMPT
+      sprintf(
+        claude_prompts[:system_prompt],
+        context: context.search_results_prompt_formatted(link_token_mapper),
+      )
+    end
+
+    def claude_prompts
+      Rails.configuration.govuk_chat_private.llm_prompts.claude.structured_answer
```
jackbot, line 60: This has been extracted into a method on `Claude.prompt_config` now, so we can use that instead.

Also I'd consider renaming this method from `claude_prompts` to `prompt_config` maybe? The word "claude" seems a bit superfluous.

jackbot (approved):
One little comment but otherwise LGTM. Don't need to re-review.

PR #42, opened 2025-02-07
Use namespaced LLM prompt config files

jackbot:
The govuk_chat_private gem has been modified to move the LLM prompts
into provider-specific namespaces, so we need to update the calls to the
config accordingly.


kevindew (approved):
Nice one

PR #45, opened 2025-02-07
Show question's answer strategy on admin show page

jackbot:
Now that we're starting to use Bedrock, it's useful when developing to
know which answer strategy was used.

Screenshot 2025-02-07 at 15 13 27


kevindew (approved):
Nice idea

PR #59, opened 2025-02-12
Pin OpenSearch to 2.18.0 for GitHub Actions

kevindew:
Since OpenSearch 2.19.0 was released yesterday we've not been able to have a successful CI run. Our tests had froze at the point in the code that interfaces with OpenSearch [1].

I've not been able to find reference to this being a confirmed OpenSearch bug, nor been able to replicate it locally with 2.19.0 so I think will pin for now to see if it's resolved soon with a future release.

[1]: https://github.com/alphagov/govuk-chat/blob/0067666dd4f271f79afc91a023d5791b0f10cb7f/spec/spec_helper.rb#L45

### DIFF
--- a/.github/workflows/rspec.yml
+++ b/.github/workflows/rspec.yml
@@ -21,7 +21,9 @@ jobs:
   run-rspec:
     services:
       opensearch-2:
-        image: opensearchproject/opensearch:2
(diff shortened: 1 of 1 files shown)


jackbot (approved):
Good sleuthing!

PR #61, opened 2025-02-18
Prevent tests leaking prompts

jackbot:
We saw an issue recently where a spec failed on CI due to a prompt being
changed in govuk_chat_private and the RSpec output leaked the entire
content of the prompt.

We explicitly want to keep the prompts private, so this shouldn't
happen.

The main culprit is when we make an HTTP request which has been stubbed
out using Webmock, but the request made doesn't match the stubbed
request. Webmock then prints out both the stubbed body and the request
body. If the spec doesn't stub the prompt loading, then the prompt from
the private gem gets printed.

Really the only way to prevent this from happening is to never allow the
prompt config to be loaded in a spec, and to stub it out instead, so
that's what this commit does.

Mostly we just load a system prompt and a user prompt from the config
files, which are relatively straightforward to stub. Usually these can
be stubbed with a simple string, but sometimes some interpolation is
needed if the prompt contains interpolation tokens (e.g. "{input}").

This means it's a bit annoying if the prompt interpolation changes -
we'd need to remember to update the specs.

A lot of this will go away once we transition away from OpenAI, because
we don't use Webmock for stubbing with Bedrock.

Nothing sensitive is automatically displayed when testing against
Bedrock. Instead we have control over what we show in the test output,
and we can hide anything sensitive from being shown on CI, as is done in
this commit.

Initially I'd wanted to try to suppress the Webmock output globally if
in CI, but it doesn't appear to be possible.

### DIFF
--- a/spec/lib/answer_composition/pipeline/claude/question_rephraser_spec.rb
+++ b/spec/lib/answer_composition/pipeline/claude/question_rephraser_spec.rb
@@ -5,7 +5,18 @@ RSpec.describe AnswerComposition::Pipeline::Claude::QuestionRephraser do
   let(:question_records) { conversation.questions.joins(:answer) }
 
   context "when there is a valid response from Claude" do
-    let(:expected_user_prompt) do
+    let(:rephrased) { "How do I pay my corporation tax" }
+
+    it "includes the current question in the user prompt" do
+      client = stub_bedrock_converse(
+        bedrock_claude_text_response(rephrased, user_message: Regexp.new(question.message)),
+      )
+      described_class.call(question.message, question_records)
+
+      expect(client.api_requests.size).to eq(1)
+    end
+
+    it "includes the message_history in the user prompt" do
       message_history = <<~HISTORY.strip
         user:
         """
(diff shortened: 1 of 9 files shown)


spec/lib/answer_composition/pipeline/claude/question_rephraser_spec.rb
```diff
@@ -5,7 +5,16 @@
   let(:question_records) { conversation.questions.joins(:answer) }
 
   context "when there is a valid response from Claude" do
-    let(:expected_user_prompt) do
+    let(:rephrased) { "How do I pay my corporation tax" }
+
+    it "includes the current question in the user prompt" do # rubocop:disable RSpec/NoExpectationExample
```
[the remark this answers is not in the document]
kevindew: Let's use the client returned by stub_bedrock_converse as something to assert on.

I'm not sure of the interface so please research it a little, but I think we can do `expect(client.api_requests.size).to be(1)` as a starter

spec/lib/answer_composition/pipeline/open_ai/question_rephraser_spec.rb
```diff
@@ -23,23 +23,22 @@
...
-        { role: "user", content: user_prompt },
-      ]
     end
     let(:rephrased) { "How do I pay my corporation tax" }
 
-    before do
-      stub_openai_chat_completion(expected_messages, answer: rephrased)
+    it "includes the current question in the user prompt" do # rubocop:disable RSpec/NoExpectationExample
+      stub_openai_chat_completion(Regexp.new(question.message))
+      described_class.call(question.message, question_records)
```
kevindew, line 31: we should also do an assertion here we can get the request as a response from the stub and do a expect(request).to have_been_requested (or something similar)

kevindew (commented):
Looks good, just a few comments on the tests without expectations - I've had to rush for a meeting though so apologies if this is an abrupt end.

spec/lib/answer_composition/pipeline/open_ai/question_rephraser_spec.rb
jackbot: Updated :+1:

kevindew (approved):
Great work

PR #72, opened 2025-02-21
Claude: question routing

jackbot:
Adds a question routing step into the pipeline used for questions using
the Claude answer strategy.

It's mostly a direct port of the OpenAI one, and behaves in exactly the
same way, i.e. creates a tool from each of the classifications in the
config and gets the LLM to use one of those tools when responding.

### DIFF
--- a/lib/answer_composition/composer.rb
+++ b/lib/answer_composition/composer.rb
@@ -55,6 +55,7 @@ module AnswerComposition
       when "claude_structured_answer"
         PipelineRunner.call(question:, pipeline: [
           Pipeline::QuestionRephraser.new(llm_provider: :claude),
+          Pipeline::Claude::QuestionRouter,
           Pipeline::SearchResultFetcher,
           Pipeline::Claude::StructuredAnswerComposer,
         ])
--- a/lib/answer_composition/pipeline/claude/question_router.rb
+++ b/lib/answer_composition/pipeline/claude/question_router.rb
@@ -1,0 +1,174 @@
+module AnswerComposition::Pipeline
+  module Claude
+    class QuestionRouter
+      BEDROCK_MODEL = "eu.anthropic.claude-3-5-sonnet-20240620-v1:0".freeze
+
+      def self.call(...) = new(...).call
+
+      def initialize(context)
+        @context = context
+      end
+
+      def call
+        start_time = Clock.monotonic_time
+
+        answer = context.answer
+        answer.assign_llm_response("question_routing", bedrock_response.to_h)
+
+        if genuine_rag?
+          answer.assign_attributes(
+            question_routing_label:,
+            question_routing_confidence_score: llm_classification_data["confidence"],
+          )
+        else
+          answer.assign_attributes(
+            message: use_llm_answer? ? llm_answer : Answer::CannedResponses.response_for_question_routing_label(question_routing_label),
+            status: answer_status,
+            question_routing_label:,
+            question_routing_confidence_score: llm_classification_data["confidence"],
+          )
+
+          context.abort_pipeline unless use_llm_answer?
+        end
+
+        answer.assign_metrics("question_routing", build_metrics(start_time))
+      end
+
+    private
+
+      attr_reader :context
+
+      def label_config
+        Rails.configuration.question_routing_labels.fetch(question_routing_label)
+      end
+
+      def use_llm_answer?
+        return false if token_limit_reached?
+
+        label_config[:use_answer]
+      end
+
+      def token_limit_reached?
+        bedrock_response["stop_reason"] == "max_tokens"
+      end
+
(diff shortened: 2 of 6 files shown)


kevindew (commented):
Just got to dash for a meeting but added some quick thoughts

lib/answer_composition/pipeline/claude/question_router.rb
```diff
@@ -0,0 +1,174 @@
...
+      end
+
+      def prompt_config
+        Claude.prompt_config.question_routing
+      end
+
+      def tool_config
+        {
+          tools:,
+          tool_choice: { any: {} },
```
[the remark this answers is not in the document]
kevindew: > I guess the primary driver for using invokeModel is that we thought we'd have more control over the tool usage, specifically the disable_parallel_tool_choice parameter, is that right? So if the converse API was using multiple tools to classify a question then we'd be able to have more control over it by using invokeModel?

Yup that's correct. It's essentially that if Converse forces multiple tools, we likely have a bug we can't fix without changing the API.

The other reasons for going down InvokeModel are:

- if we get bugs with Converse - we were warned that there are occassionally inconsistencies that manifest as issues that need fixing on the converse side
- for future expected needs, but we don't need to worry about that yet. Particularly having a response return citations

All good to stick with converse until we know if we've got a bug with parallel tool calls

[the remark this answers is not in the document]
jackbot: Cool, cheers for that.

In that case I've fixed the pipeline ordering issue here, so would you mind doing another review if we're happy to stick to `converse`?

I've pushed a branch ([`claude-question-router-invoke-api`](https://github.com/alphagov/govuk-chat/tree/claude-question-router-invoke-api)) which has the application code changes for switching the question router to use `invokeModel`. I've not updated the specs but it'll serve as a reference for the future of how to use the `invokeModel` API, as it's a bit different to `converse`.

lib/answer_composition/composer.rb
```diff
@@ -56,6 +56,7 @@ def compose_answer
         PipelineRunner.call(question:, pipeline: [
           Pipeline::QuestionRephraser.new(llm_provider: :claude),
           Pipeline::SearchResultFetcher,
+          Pipeline::Claude::QuestionRouter,
```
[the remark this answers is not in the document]
jackbot: Oops, good catch!

lib/answer_composition/pipeline/claude/question_router.rb
```diff
@@ -0,0 +1,174 @@
...
+          model_id: BEDROCK_MODEL,
+          messages:,
+          inference_config:,
+          tool_config:,
+        )
+      end
+
+      def inference_config
+        {
+          max_tokens: 4096,
```
kevindew, line 97: We have this at a very low value for OpenAI to remove the risk of the prompt being output verbatim: https://github.com/alphagov/govuk-chat/blob/adb8eb27266d7f6979449df15ece3202a5f88b7b/lib/answer_composition/pipeline/openai/question_router.rb#L4

spec/support/stub_bedrock.rb
```diff
@@ -39,7 +39,24 @@ def bedrock_claude_structured_answer_response(question, answer)
...
+
+  def bedrock_claude_question_routing_response(question)
+    lambda do |context|
+      given_question = context.params.dig(:messages, -1, :content, 0, :text)
+
+      if question && given_question != question
+        raise "Unexpected question received: \"#{given_question}\". Expected \"#{question}\"."
+      end
+
+      bedrock_claude_response("Test")
```
kevindew, line 54: Is this something left behind?

spec/lib/answer_composition/pipeline/claude/question_router_spec.rb
```diff
@@ -0,0 +1,240 @@
...
+    [
+      {
+        tool_spec: {
+          name: "greetings",
+          description: "A classification description",
+          input_schema: {
+            json: {
+              type: "object",
+              properties: properties.merge({
+                confidence: llm_prompts[:confidence_property],
```
kevindew, line 31: I wonder if rather than having `llm_prompts` as a method that is only used for this it is better to inline it as `confidence_property = Rails.configuration.govuk_chat_private.llm_prompts[:confidence_property]` where the properties variable is defined.

kevindew (commented):
Great job, just a few small things here but otherwise looks just about ready to merge.

lib/answer_composition/pipeline/claude/question_router.rb
jackbot: I've updated Claude to the same number of tokens and added a comment as to why it's lower.

kevindew (approved):
Nice one 👍

PR #79, opened 2025-02-28
Claude: Output Guardrails

chaecramb:
### Add Claude model support for output guardrails

This PR adds support for using Claude models in our output guardrails system, previously only available for OpenAI models. It implements the approach outlined in the [card](https://trello.com/c/qWFIiUL5/2303-port-the-output-guardrails-for-claude-bedrock), making the guardrails system model-agnostic by allowing specification of an LLM provider.

## Changes

- Refactored the guardrails system to use a provider-based approach, similar to `QuestionRephraser`
- **New components**: 
  - Added `Guardrails::Claude::MultipleChecker` to handle Claude-specific implementation
  - Added `Guardrails::OpenAI::MultipleChecker` containing the existing OpenAI implementation
  - Modified base `MultipleChecker` to delegate to the appropriate provider implementation
- **Pipeline updates**:
  - Added Claude guardrails to the Claude pipeline
  - Updated `OutputGuardrails` to accept an `llm_provider` parameter
  - Modified both `AnswerGuardrails` and `QuestionRoutingGuardrails` to work with the new architecture


lib/answer_composition/pipeline/answer_guardrails.rb
```diff
@@ -1,7 +1,9 @@
 module AnswerComposition
   module Pipeline
     class AnswerGuardrails < OutputGuardrails
-      def call
+      def call(context)
+        @context = context
```
kevindew, line 5:
```suggestion
```

Would suggest not setting this an ivar. It's not particularly idiomatic to set ivars outside of initialize methods and as this isn't used outside the `#call` method we don't gain anything from it.

lib/answer_composition/pipeline/answer_guardrails.rb
```diff
@@ -19,6 +21,10 @@ def call
       rescue ::Guardrails::MultipleChecker::ResponseError => e
         abort_after_response_error(e, start_time, Answer::CannedResponses::ANSWER_GUARDRAILS_FAILED_MESSAGE)
       end
+
+    private
+
+      attr_reader :context
```
kevindew, line 27:
```suggestion

    private

      attr_reader :context
```

lib/answer_composition/pipeline/question_routing_guardrails.rb
```diff
@@ -1,7 +1,8 @@
 module AnswerComposition
   module Pipeline
     class QuestionRoutingGuardrails < OutputGuardrails
-      def call
+      def call(context)
+        @context = context
```
kevindew, line 5:
```suggestion
```

As per previous comment

lib/answer_composition/pipeline/question_routing_guardrails.rb
```diff
@@ -21,6 +22,10 @@ def call
       rescue ::Guardrails::MultipleChecker::ResponseError => e
         abort_after_response_error(e, start_time, Answer::CannedResponses::QUESTION_ROUTING_GUARDRAILS_FAILED_MESSAGE)
       end
+
+    private
+
+      attr_reader :context
```
kevindew, line 28:
```suggestion
```

lib/guardrails/multiple_checker.rb
```diff
@@ -142,5 +121,17 @@ def extract_guardrails(parts)
...
+      all_guardrail_numbers = guardrail_numbers.map(&:to_s).join(", ")
+      longest_possible_response_string = %(True | "#{all_guardrail_numbers}")
+
+      token_count = Tiktoken
+       .encoding_for_model(::Guardrails::OpenAI::MultipleChecker::OPENAI_MODEL)
+       .encode(longest_possible_response_string)
+       .length
+
+      token_count + MAX_TOKENS_BUFFER
+    end
```
kevindew, line 135: Hmm I guess we may have to rethink this a bit since it is based on OpenAI tokens

kevindew (commented):
Just put a couple of quick comments in before running to a meeting - I've not had a chance yet to look at much of it. Thought it might be worth dropping a comment though so we could get thinking about the tiktoken issue.

lib/answer_composition/pipeline/answer_guardrails.rb
chaecramb: I'm scratching my head as to why I added this now, since it's so obviously redundant. I'll remove it.

chaecramb: Ah, wait a minute, having seen the tests blow up after removing it, I now remember. It's accessed via the `attr_reader` here https://github.com/alphagov/govuk-chat/blob/b01049125a60739652e375f5f2fc5eec0102c87c/lib/answer_composition/pipeline/output_guardrails.rb#L32

But yeah, there's presumably a better pattern I can use here, as it looks totally out of place when looking at answer guardrails and question routing guardrails in isolation.

chaecramb: Would using `self.context = context` in `AnswerGuardrails` and `QuestionRoutingGuardrails`. Then putting an `attr_accessor` for it in `OutputGuardrails` be more idiomatic?

lib/guardrails/multiple_checker.rb
chaecramb: We could potentially create an encoding with Tiktoken that matches Claude's, as shown here: https://github.com/IAPark/tiktoken_ruby/issues/15

chaecramb: Actually, it looks like `tiktoken_ruby` can't configure the encoding using json as demonstrated in the Python version.

lib/answer_composition/pipeline/output_guardrails.rb
```diff
@@ -1,28 +1,35 @@
...
+            llm_prompt_tokens: response_or_error.llm_token_usage["prompt_tokens"],
+            llm_completion_tokens: response_or_error.llm_token_usage["completion_tokens"],
+            llm_cached_tokens: response_or_error.llm_token_usage.dig("prompt_tokens_details", "cached_tokens"),
+          }
+        when :claude
+          {
+            duration: Clock.monotonic_time - start_time,
+            llm_prompt_tokens: response_or_error.llm_token_usage["input_tokens"],
+            llm_completion_tokens: response_or_error.llm_token_usage["output_tokens"],
+          }
```
kevindew, line 26: I think we should try avoid this case statement as it feels like the response_or_error should be able to handle the abstraction.

What I'd suggest doing is adjusting the response_or_error classes so they have attributes for llm_prompt_tokens, llm_completion_tokens and llm_cached_tokens and then this code can be:

```
        {
          duration: Clock.monotonic_time - start_time,
          llm_prompt_tokens: response_or_error.llm_prompt_tokens,
          llm_completion_tokens: response_or_error.llm_completion_tokens,
          llm_cached_tokens: response_or_error.llm_cached_tokens,
        }
```

lib/guardrails/claude/multiple_checker.rb
```diff
@@ -0,0 +1,41 @@
+module Guardrails
+  module Claude
+    class MultipleChecker
+      BEDROCK_MODEL = "eu.anthropic.claude-3-5-sonnet-20240620-v1:0".freeze
+      MAX_TOKENS_BUFFER = 5
```
kevindew, line 5: I don't think this constant is used?

lib/tasks/guardrails.rake
```diff
@@ -59,7 +59,7 @@ namespace "guardrails" do
       abort("Invalid guardrail type. Valid guardrail types are #{valid_guardrail_types.to_sentence}")
     end
 
-    prompt = Guardrails::MultipleChecker.collated_prompts(guardrail_type)
+    prompt = Guardrails::MultipleChecker.collated_prompts(guardrail_type, :openai)
```
kevindew, line 62: Rather than this rake task being specific for OpenAI we should make it so it can take a llm_provider attribute so that the rake task can be run for openai and claude so we can compare results.

lib/tasks/guardrails.rake
```diff
@@ -19,13 +19,13 @@ namespace "guardrails" do
 
     output_path = args[:output_path]
 
-    model_name = Guardrails::MultipleChecker::OPENAI_MODEL
+    model_name = Guardrails::OpenAI::MultipleChecker::OPENAI_MODEL
```
kevindew, line 22: I think we should remove this usage of the model name being included in the results, it seems inconsistently precise compared to the rest of the code which is an abstraction.

lib/guardrails/openai/multiple_checker.rb
```diff
@@ -0,0 +1,50 @@
...
+      def call
+        llm_response = openai_response.dig("choices", 0)
+        llm_guardrail_result = llm_response.dig("message", "content")
+        llm_token_usage = openai_response["usage"]
+
+        {
+          llm_response:,
+          llm_guardrail_result:,
+          llm_token_usage:,
+        }
```
kevindew, line 24: Very minor but you can save a few lines here:

```suggestion
        {
          llm_response: openai_response.dig("choices", 0)
          llm_guardrail_result: llm_response.dig("message", "content")
          llm_token_usage: openai_response["usage"],
        }
```

lib/guardrails/multiple_checker.rb
kevindew: Sure, I think we may just want to do something a bit more basic and see if any problems occur.

I'd suggest we just hardcode both of the dedicated classes to have a max tokens of 100 which should give up to about 50 guardrails without a problem - and I believe we'd expect to have some real other problems if we had 50 guardrails.

spec/support/guardrails_examples.rb
```diff
@@ -1,23 +1,23 @@
 module GuardrailsExamples
   shared_examples "a passing guardrail pipeline step" do |guardrail_name|
     it "calls the guardrails with the answer message" do
-      described_class.call(context)
+      described_class.new(llm_provider: :openai).call(context)
```
kevindew, line 4: I imagine we could benefit from setting a default parameter in the classes for this (:openai) so we don't have lots of repetition of this argument which doesn't impact the test.

(for context I was wondering should we have claude tests too but looked into it and it shouldn't matter because of the mocking)

kevindew (commented):
Nice job Chae, I've added a few comments in with suggestions - looks like this isn't far off, with maybe some digging still needed around the rake task.

I did wonder whether we wanted to reduce some of the duplications between the specs for Guardrails::Claude::MultipleChecker and Guardrails::OpenAI::MultipleChecker, but I think they're ok since it's only 2 files and it's very unclear if there'll be a third.

lib/tasks/guardrails.rake
chaecramb: My intention was to switch on the provider:

```
    model_name = case llm_provider
                 when :openai
                   Guardrails::OpenAI::MultipleChecker::OPENAI_MODEL
                 when :claude
                   Guardrails::Claude::MultipleChecker::BEDROCK_MODEL
                 end
```

Do you think we just don't need the model name in the output at all?

lib/answer_composition/pipeline/question_routing_guardrails.rb
kevindew: Should now be able to remove this

lib/answer_composition/pipeline/question_routing_guardrails.rb
kevindew: This also should be fine for removal now

lib/tasks/guardrails.rake
kevindew: Yeah I don't think we need it. Feels quite an internal detail.

kevindew (approved):
Great stuff

PR #110, opened 2025-03-19
Claude jailbreak guardrails

chaecramb:
This PR adds support for Claude-based jailbreak detection, allowing us to port the jailbreak guardrails functionality to AWS Bedrock. It completes the process of porting our LLM functionality to AWS (with just OpenAI embeddings remaining).

## Changes
- Extracted OpenAI-specific jailbreak checking logic into its own module (`Guardrails::OpenAI::JailbreakChecker`)
- Created a new Claude-specific jailbreak checker (`Guardrails::Claude::JailbreakChecker`)
- Refactored the main `JailbreakChecker` class to delegate to provider-specific implementations
- Updated the `JailbreakGuardrails` pipeline class to accept an LLM provider parameter
- Added Claude as a provider option in the pipeline
- Updated the jailbreak response evaluation task to accept Claude as a provider
- Added tests for the new functionality

## Implementation Details
- The Claude jailbreak checker currently uses AWS Bedrock's Claude 3.5 Sonnet model
- The implementation follows the same pattern as our output guardrails
- The main `JailbreakChecker` handles common logic like interpreting results and creating the final response object
- Provider-specific classes handle the actual API calls and formatting of results

## Testing
- Unit tests have been added for the Claude implementation
- The evaluation rake task can now be run with Claude as a provider:
  ```
  INPUT="your test input" 
  bundle exec rake evaluation:jailbreak_response[claude]
  ```

Closes: https://trello.com/c/9EwyIVki/2330-port-the-jailbreak-guardrails-for-claude-bedrock

### DIFF
--- a/spec/lib/answer_composition/composer_spec.rb
+++ b/spec/lib/answer_composition/composer_spec.rb
@@ -45,6 +45,7 @@ RSpec.describe AnswerComposition::Composer do
         stub_pipeline_initialize(AnswerComposition::Pipeline::QuestionRephraser, llm_provider: :openai)
         stub_pipeline_initialize(AnswerComposition::Pipeline::QuestionRoutingGuardrails, llm_provider: :openai)
         stub_pipeline_initialize(AnswerComposition::Pipeline::AnswerGuardrails, llm_provider: :openai)
+        stub_pipeline_initialize(AnswerComposition::Pipeline::JailbreakGuardrails, llm_provider: :openai)
 
         expected_pipeline = [
           AnswerComposition::Pipeline::JailbreakGuardrails.new(llm_provider: :openai),
@@ -75,6 +76,7 @@ RSpec.describe AnswerComposition::Composer do
         stub_pipeline_initialize(AnswerComposition::Pipeline::QuestionRephraser, llm_provider: :claude)
         stub_pipeline_initialize(AnswerComposition::Pipeline::QuestionRoutingGuardrails, llm_provider: :claude)
         stub_pipeline_initialize(AnswerComposition::Pipeline::AnswerGuardrails, llm_provider: :claude)
+        stub_pipeline_initialize(AnswerComposition::Pipeline::JailbreakGuardrails, llm_provider: :claude)
 
         expected_pipeline = [
           AnswerComposition::Pipeline::JailbreakGuardrails.new(llm_provider: :claude),
--- a/spec/lib/answer_composition/composer_spec.rb
+++ b/spec/lib/answer_composition/composer_spec.rb
@@ -47,7 +47,7 @@ RSpec.describe AnswerComposition::Composer do
         stub_pipeline_initialize(AnswerComposition::Pipeline::AnswerGuardrails, llm_provider: :openai)
 
         expected_pipeline = [
-          AnswerComposition::Pipeline::JailbreakGuardrails,
+          AnswerComposition::Pipeline::JailbreakGuardrails.new(llm_provider: :openai),
           AnswerComposition::Pipeline::QuestionRephraser.new(llm_provider: :openai),
           AnswerComposition::Pipeline::OpenAI::QuestionRouter,
           AnswerComposition::Pipeline::QuestionRoutingGuardrails.new(llm_provider: :openai),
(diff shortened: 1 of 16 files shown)


spec/lib/answer_composition/composer_spec.rb
```diff
@@ -39,6 +39,9 @@
         rephraser = instance_double(AnswerComposition::Pipeline::QuestionRephraser)
         allow(AnswerComposition::Pipeline::QuestionRephraser)
           .to receive(:new).with(llm_provider: :openai).and_return(rephraser)
+        jailbreak_guardrails = instance_double(AnswerComposition::Pipeline::JailbreakGuardrails)
+        allow(AnswerComposition::Pipeline::JailbreakGuardrails)
+          .to receive(:new).with(llm_provider: :openai).and_return(jailbreak_guardrails)
```
kevindew, line 44: I imagine we could benefit from cleaning these up with some sort of helper method - as we have quite a lot of boiler plate in this file. Fine to do as a follow up task (I suppose I could open a quick PR myself - I'll do that)

Something like:

```
jailbreak_guardrails = stub_pipeline_instance(AnswerComposition::Pipeline::JailbreakGuardrails, llm_provider: :openai)

def stub_pipeline_instance(klass, *args, **kwargs)
  double = instance_double(klass)
  allow(klass).to receive(:new).with(*args, **kwargs).and_return(double)
end

File: spec/lib/guardrails/claude/jailbreak_checker_spec.rb [hunk 2]
```diff
@@ -0,0 +1,19 @@
+RSpec.describe Guardrails::Claude::JailbreakChecker do
```
kevindew, line 1: We should have the test within a `describe ".call"` block

spec/support/stub_openai_chat.rb
```diff
@@ -132,20 +132,24 @@ def stub_openai_output_guardrail(to_check, response = "False | None")
   end
 
   def stub_openai_jailbreak_guardrails(to_check, response = "PassValue")
```
kevindew, line 134: Should this be also changed to have a triggered option rather than the exact response?

kevindew (commented):
Looks great - just a few minor comments. I'll quickly open a PR to try reduce boilerplate on the pipeline stubbing but don't worry about waiting for that - it can go in after yours

spec/lib/answer_composition/composer_spec.rb
kevindew: I've opened: https://github.com/alphagov/govuk-chat/pull/117

spec/support/stub_openai_chat.rb
chaecramb: Yeah, that would keep it consistent with the bedrock approach, and is simpler. The only thing is we currently have this test for an unknown response from OpenAI, which become more complicated to re-create. What do you think?

```
  it "raises a response error when the LLM returns a different response" do
    stub_openai_jailbreak_guardrails(input, "?")
    expected_error = an_instance_of(described_class::ResponseError).and(
      having_attributes(
        message: "Error parsing jailbreak guardrails response",
        llm_guardrail_result: "?",
        llm_response: hash_including("message", "finish_reason", "index"),
        llm_prompt_tokens: be_a(Integer),
        llm_completion_tokens: be_a(Integer),
        llm_cached_tokens: be_a(Integer).or(be_nil),
      ),
    )

    expect { described_class.call(input) }
      .to raise_error(expected_error)
  end
  ```

Comment (chaecramb) [reply spec/support/stub_openai_chat.rb:134]: A little more complex, but I could do:

```
def stub_openai_jailbreak_guardrails(to_check, triggered: false, override_response: nil)
...
  response = if override_response
               override_response
             else
               triggered ? "FailValue" : "PassValue"
             end
...
```

Not sure if that's worth it though.

Comment (kevindew) [reply spec/support/stub_openai_chat.rb:134]: Thanks, so I don't think we should do the override suggestion there as it seems too niche. I think instead we've got two options:

1) mock chat-completion 
2) don't actually hit an LLM and stub.

I think the second is a better option since this classes unit isn't the direct LLM interaction. So something like:

```
allow(OpenAI::JailbreakChecker.call).to receive(question).and_return( { 
  llm_response: {},
  llm_guardrail_result: "unexpected,
  llm_prompt_tokens: 10,
  llm_completion_tokens: 10,
  llm_cached_tokens: nil,
}
```

PR #117, opened 2025-03-21
Reduce boilerplate in stubbing pipelines

kevindew:
This adds a helper function that can set up an instance double and a stubbing of the new method on classes.

It allows reducing the rather terse code that is used to the stubbing multiple times.


chaecramb (approved):
Looks good

PR #139, opened 2025-04-01
Add a rake task to fetch output guardrails results

chaecramb:
Trello: https://trello.com/c/0ssf3xuh/2376-port-output-guardrail-evaluations-to-the-python-evaluation-framework

This task is part of the autoevaluation pipeline. When given an input, a provider and a guardrail type it outputs the result of the call to the specified output guardrail.


lib/guardrails/multiple_checker.rb
```diff
@@ -77,6 +77,8 @@ def call
         response = OpenAI::MultipleChecker.call(input, prompt)
       when :claude
         response = Claude::MultipleChecker.call(input, prompt)
+      else
+        raise "Unsupported provider: #{llm_provider}"
```
kevindew, line 81: This looks good, but expect we want it covered by a unit test for multiple checker

kevindew (commented):
Looks great, just a small change needed

spec/lib/tasks/evaluation_spec.rb
```diff
@@ -127,4 +127,61 @@
...
+    let(:input) { "input" }
+
+    before { Rake::Task[task_name].reenable }
+
+    it "requires a INPUT env var" do
+      expect { Rake::Task[task_name].invoke("openai", "answer_guardrails") }
+        .to raise_error("Requires an INPUT env var")
+    end
+
+    it "requires a provider" do
```
jackbot, line 142: I'd say this would be better covered by a unit test of the `lib/guardrails/multiple_checker.rb` class rather than an integration test

spec/lib/tasks/evaluation_spec.rb
```diff
@@ -127,4 +127,61 @@
...
+    end
+
+    it "requires a guardrail type" do
+      ClimateControl.modify(INPUT: input) do
+        expect { Rake::Task[task_name].invoke("openai") }
+          .to raise_error("Requires a guardrail type")
+      end
+    end
+
+    it "requires a known guardrail type" do
```
jackbot, line 163: This logic is already tested in the `multiple_checker_spec` files for each provider, so probably no point in duplicating it here.

spec/lib/tasks/evaluation_spec.rb
```diff
@@ -127,4 +127,61 @@
...
+      ClimateControl.modify(INPUT: input) do
+        expect { Rake::Task[task_name].invoke("openai", "unknown") }
+          .to raise_error("No LLM prompts found for unknown")
+      end
+    end
+
+    it "outputs the response as JSON to stdout" do
+      ClimateControl.modify(INPUT: input) do
+        result = Guardrails::MultipleChecker::Result.new(
```
jackbot, line 172: Wonder if you could use the existing factory (`spec/factories/output_guardrail_result_factory.rb`) here?

spec/lib/tasks/evaluation_spec.rb
chaecramb: Just to double check, was this comment supposed to be "requires a known provider": https://github.com/alphagov/govuk-chat/blob/3c1c2723c74817c611a7f3a46c9e9cbb66d0e49d/spec/lib/tasks/evaluation_spec.rb#L149

jackbot: Yeah sorry, commented on the wrong spec here. It's referring to the "requires a known provider" test.

chaecramb:
@kevindew, @jackbot comments have been addressed

jackbot (approved):
LGTM

kevindew (approved):
Thanks, looks good to me too 👍

PR #142, opened 2025-04-03
Rake task to generate a RAG answer

kevindew:
Trello: https://trello.com/c/aCMTRtwW/2380-dev-and-data-science-work-together-to-write-evaluation-code-for-rag-questions

Data science are keen to have a means to evaluate how GOV.UK Chat performs at the RAG level without the other components such as guardrails and question routing.

To facilitate this I have written this Rake task which is expected to be called from the Python evaluation framework [1] as a means to generate the answers.

There was some key considerations in putting this together:

1. Where to put the code for generating the answer

I considered hardcoding the answer generation or
creating new answer strategies as means to implement this. I chose the approach of hardcoding as I felt it offered a route to be less involved in the AnswerComposition::Composer code and allowed skipping forbidden words and other error handling.

Doing it this approach has probably increased the risk that this code could be less maintained than the answer strategy since it is now in a potentially surprisingly place.

2. Raising an error if an answer has an error status

For evaluation we ideally want all the questions to be answered. There are however a couple of ways when they may not be answered - abort statuses (such as unanswerable_no_govuk_content) and error statuses (such as error_answer_service_error).

I decided to return data when it is an abort status as, following a chat with Alessia, it looked like it'd be good to have that as data for a report even though these wouldn't be explicitly evaluated.

Whereas for errors I thought it was best to raise here to make very clear something had gone wrong. There's definitely room for improvement with the error handling as just blowing up is a blunt instrument. I do expect however that the majority of errors will be lack of an env var and things that would indicate the whole process would fail so I think erroring early is best.

3. Returning answer data via to_json

This decision is unlikely to be sufficient for long (we'll need extra data like sources) but struck me as the simplest way to get data out that we could use in evaluation quickly.

I expect this to be iterated.

[1]: https://github.com/alphagov/govuk-chat-evaluation-prototype/pull/12


lib/tasks/evaluation.rake
```diff
@@ -64,4 +64,30 @@ namespace :evaluation do
 
     puts(response.to_json)
   end
+
+  desc "Produce the output of a RAG response for a user input"
+  task :generate_rag_structured_answer_response, %i[provider] => :environment do |_, args|
+    raise "Requires an INPUT env var" if ENV["INPUT"].blank?
```
davidgisbey, line 70: def not a blocker. Just wondering why we're using an env var for the input and not an arg? i.e.

`task :generate_rag_structured_answer_response, %i[provider question]`

Is that a convention that GOV.UK apps tend not to adhere to? I can see we've also used that in the evaluation rake task.

davidgisbey (approved):
Looks good to me. I've left a question on why we use an env var for the question message but def not a blocker 👍

lib/tasks/evaluation.rake
```diff
@@ -64,4 +64,30 @@ namespace :evaluation do
...
+    raise "Requires a provider" if args[:provider].blank?
+
+    question = Question.new(message: ENV["INPUT"], conversation: Conversation.new)
+    answer = case args[:provider]
+             when "openai"
```
davidgisbey, line 75: seems totally reasonable to hardcode these here to me rather than going through the faff of creating an answer strategy. If the pipeline gets massively changed and these fall out of sync we can always revisit it down the line.

lib/tasks/evaluation.rake
kevindew: Yeah, it's not an area I feel like I know absolute rules for it. In this task I just followed the existing convention, but as I set-up the earlier convention I should also justify that.

Rake arguments in general are a bit of a pain once they're beyond a single string and I felt with questions being potentially complex strings with potentially commas and quotes inside them that we side step a lot of usability issues by using an env var.

davidgisbey: Okay cool makes sense to me 👍

PR #145, opened 2025-04-07
Evaluation rake task: add question routing

jackbot:
https://trello.com/c/82aouRre/2377

Adds a new task to generate a question routing label for a given input.

### DIFF
--- a/spec/lib/tasks/evaluation_spec.rb
+++ b/spec/lib/tasks/evaluation_spec.rb
@@ -1,4 +1,27 @@
 RSpec.describe "rake evaluation tasks" do
+  shared_examples "a task requiring input and provider" do
+    it "requires a INPUT env var" do
+      expect { Rake::Task[task_name].invoke("openai") }
+        .to raise_error("Requires an INPUT env var")
+    end
+
+    it "requires a provider" do
+      ClimateControl.modify(INPUT: input) do
+        expect { Rake::Task[task_name].invoke }
+          .to raise_error("Requires a provider")
+      end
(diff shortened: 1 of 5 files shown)


spec/lib/tasks/evaluation_spec.rb
```diff
@@ -1,4 +1,28 @@
...
+      ClimateControl.modify(INPUT: input) do
+        expect { Rake::Task[task_name].invoke }
+          .to raise_error("Requires a provider")
+      end
+    end
+  end
+
+  shared_examples "a task requiring a known provider" do |error_message|
+    it "raises an error when given an unknown provider" do
+      msg = error_message || "Unexpected provider super-duper-ai"
```
davidgisbey, line 18: Minor nitty comment so feel free to ignore. I wonder if we should just update `Unsupported` to `Unexpected` or vice versa?

Then it could just match on`expect { Rake::Task[task_name].invoke("super-duper-ai") }
          .to raise_error(/Unexpected provider/)` and it wouldn't need the conditional arg.

Unless they're subtly different and i'm missing something.

davidgisbey (approved):
i left a v minor nitty comment. Feel free to ignore as it's def not a blocker

spec/lib/tasks/evaluation_spec.rb
jackbot: Yep good shout, I've made that change.

davidgisbey (approved):
Looks awesome. Thanks for that 🚢

PR #153, opened 2025-04-08
[SPIKE] Add blueprinter gem and use it to build API endpoint

davidgisbey:
## Description 

I thought i'd spike out using blueprinter for serialisation so we could discuss whether we want to use it.

This adds blueprints for the following models:

- AnswerSource
- Answer
- Question
- Conversation

Then adds a conversation show endpoint that returns a conversation with it's nested associations as JSON using ConversationBlueprint.

## Thoughts

- It was really easy and intuitive to use. The docs are great
- The `#view` functionality was really useful for being able to define the different types of questions there are (answered and pending)
- While there does seem to be a convention for the blueprints to sit in `app/blueprints`, If we want we can do `class AnswerSourceSerializer < Blueprinter::Base` and move stuff to `app/serializers` folder. It works fine.

For what we need it for which is a pretty basic use case it seems reasonable to use it to me. It'd be far easier than maintaining our own PORO serialisers we were to build them.

I've had a look into https://github.com/rails-api/active_model_serializers. There seem to be quite a few issues with limited support, release frequency and performance with many people recommending `blueprinter` and `jsonapi-serializer ` instead.

### Answered question

image

### With feedback and no sources

image

## Pending question 

image

### Pending and answered question 

image

### DIFF
--- a/app/blueprints/conversation_blueprint.rb
+++ b/app/blueprints/conversation_blueprint.rb
@@ -1,0 +1,16 @@
+class ConversationBlueprint < Blueprinter::Base
+  identifier :id
+
+  field :created_at do |conversation, _options|
+    conversation.created_at.iso8601
+  end
+
+  association :answered_questions, blueprint: QuestionBlueprint, view: :answered do |_conversation, options|
+    options[:answered_questions] || []
+  end
+
+  association :pending_question, blueprint: QuestionBlueprint, view: :pending,
+                                 if: ->(_field_name, _conversation, options) { options[:pending_question].present? } do |_conversation, options|
+    options[:pending_question]
+  end
+end
--- a/app/controllers/app_api/v0/conversations_controller.rb
+++ b/app/controllers/app_api/v0/conversations_controller.rb
@@ -1,9 +1,0 @@
-class AppApi::V0::ConversationsController < ApplicationController
-  def show
-    conversation = Conversation.includes(questions: { answer: %i[sources feedback] })
-                                .find(params[:id])
-    render json: ConversationBlueprint.render(conversation), status: :ok
-  rescue ActiveRecord::RecordNotFound
-    render json: { error: "Conversation not found" }, status: :not_found
-  end
-end
--- a/app/controllers/app_api/v0/conversations_controller.rb
+++ b/app/controllers/app_api/v0/conversations_controller.rb
@@ -1,0 +1,9 @@
+class AppApi::V0::ConversationsController < ApplicationController
+  def show
+    conversation = Conversation.includes(questions: { answer: %i[sources feedback] })
+                                .find(params[:id])
+    render json: ConversationBlueprint.render(conversation), status: :ok
+  rescue ActiveRecord::RecordNotFound
+    render json: { error: "Conversation not found" }, status: :not_found
+  end
+end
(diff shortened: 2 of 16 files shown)


config/routes.rb
```diff
@@ -143,6 +143,10 @@
     end
   end
 
+  scope :app_api, format: false, defaults: { format: "json" } do
```
kevindew, line 146: Let's just scope this as api rather than App API (we may have other clients - potentially the web version could end up being a client)

kevindew: I imagine you'll also find using a namespace better than a scope: https://guides.rubyonrails.org/routing.html#controller-namespaces-and-routing

config/routes.rb
```diff
@@ -143,6 +143,10 @@
...
+    get "/v0/conversations/:id", to: "app_api/v0/conversations#show", as: :app_api_show_conversation
```
kevindew, line 147: Let's put v0 as a scope or namespace too.

I wonder if conversation could also be a scope too for this to be the conversation API - not so clear on that though

app/blueprints/conversation_blueprint.rb
```diff
@@ -0,0 +1,13 @@
+class ConversationBlueprint < Blueprinter::Base
+  identifier :id
+
+  fields :created_at
+
+  association :answered_questions, blueprint: QuestionBlueprint, view: :answered do |conversation|
+    conversation.questions.joins(:answer)
```
kevindew, line 7: I'd expect if it'd be cleaner to have this data injected into the blueprint via attributes rather than doing the query within the serialization layer.

app/controllers/app_api/v0/conversations_controller.rb
```diff
@@ -0,0 +1,9 @@
+class AppApi::V0::ConversationsController < ApplicationController
+  def show
+    conversation = Conversation.includes(questions: { answer: %i[sources feedback] })
+                                .find(params[:id])
+    render json: ConversationBlueprint.render(conversation), status: :ok
+  rescue ActiveRecord::RecordNotFound
+    render json: { error: "Conversation not found" }, status: :not_found
```
kevindew, line 7: May well want to create blueprints for error handling too

app/blueprints/conversation_blueprint.rb
kevindew: Feels like this is doing some of the controller code rather than view if you know what I mean?

kevindew (commented):
Great work here, I'm glad this gem looks promising. I've added a few quick comments.

app/blueprints/conversation_blueprint.rb
davidgisbey: cool have passed them in at controller level 👍

config/routes.rb
```diff
@@ -143,6 +143,14 @@
     end
   end
 
+  namespace :api, format: false, defaults: { format: "json" } do
+    namespace :conversations do
+      scope :v0 do
```
kevindew, line 148: Let's go the other way round and have coversations as a namespace in v0 (sorry I probably added this ambiguity)

kevindew: Might be a dumb idea of mine too for conversations to be a namespace if we may only have one controller. Maybe scratch that idea.

kevindew: Yup API v0 with a conversations endpoint

davidgisbey:
closing this as it's been implemented

PR #162, opened 2025-04-11
Add confidence score to question routing eval output

jackbot:
We want to use the confidence score in the evaluation app, so we'll need
to output it here.

Also changed the key for the label to "classification" as this matches
the terminology used in the existing evaluation projects.


chaecramb (approved):
LGTM

PR #167, opened 2025-04-15
Add ADR for Serialization and Schema Validation in GOV.UK Chat API

davidgisbey:
## Description

This documents the findings of the spike on serialization and schema validation chat API. It documents that:

- We've decided to use Blueprinter for serializing API responses due to its simplicity, maintainability, and active support
- We chose Committee middleware for runtime validation of API responses against the OpenAPI spec, with additional validation in request specs
- That validation and error handling will remain in controllers for requests

## Relevant spikes 

#153 - spikes out adding [Blueprinter](https://github.com/procore-oss/blueprinter/blob/main/README.md) for serialisation
#163 - spikes out adding [Committee](https://github.com/interagent/committee) for schema validation

### DIFF
--- a/docs/adr/0007-serialisation-and-schema-validation-for-govuk-chat-api.md
+++ b/docs/adr/0007-serialisation-and-schema-validation-for-govuk-chat-api.md
@@ -1,0 +1,52 @@
+# 7. Serialisation and Schema Validation for GOV.UK Chat API
+
+**Date:** 2025-04-15
+
+## Context
+
+We are building an API for GOV.UK Chat, and as part of that we need to choose tools for two concerns:
+
+1. Serialising Ruby objects into JSON responses
+2. Validating API requests and responses against our OpenAPI specification
+
+These tools should be simple, reliable, and maintainable, with minimal dependencies and good community support.
+
+## Decision
+
+### Serialisation
+
+We have chosen to use [Blueprinter](https://github.com/procore/blueprinter) for serializing objects into JSON. It provides a clean and lightweight way to define how data is exposed through the API.
+
+We decided against
+
+- **active_model_serializers**, as there are issues with maintenance and it has a history of stability issues. See:
+  - https://github.com/rails-api/active_model_serializers/issues
+  - https://github.com/rails-api/active_model_serializers/issues/2396
+- **jsonapi-serializer**, We felt that our returned data models would be harder to work with were they to follow the requirements of the JSON API specification and thus provide a reduced user experience.
+
+Blueprinter is preferred because:
+
+- It’s simple to configure and easy to read
+- It keeps response logic isolated and maintainable
+- It is actively maintained and stable
+
+### OpenAPI Request and Response Validation
+
+We are using [Committee](https://github.com/interagent/committee) middleware to validate that API request and response payloads match the OpenAPI 3 specification at runtime. This gives us immediate feedback if responses drift from the specification.
+
+In addition to runtime validation, we’re also using Committee in our request specs to ensure we're checking responses conform to the schema during automated testing.
+
+We considered [rswag](https://github.com/rswag/rswag), but opted not to use it due to:
+
+- Its focus on documentation and Swagger UI, which isn’t a priority right now
+- Its more complex setup and less intuitive behaviour for schema validation
+- Committee providing simpler, targeted validation of actual responses
+
+## Status
+
+**accepted**
+
+## Consequences
+
+- Developers will use Blueprinter to define API response formats
+- Request and response validation against the OpenAPI spec will happen at runtime via middleware, and in automated tests


docs/adr/0007-serialisation-and-schema-validation-for-govuk-chat-api.md
```diff
@@ -0,0 +1,57 @@
...
+- It keeps response logic isolated and maintainable
+- It is actively maintained and stable
+
+### OpenAPI Response Validation
+
+We are using [Committee](https://github.com/interagent/committee) middleware to validate that API responses match the OpenAPI 3 specification at runtime. This gives us immediate feedback if responses drift from the specification.
+
+In addition to runtime validation, we’re also using Committee in our request specs to ensure we're checking responses conform to the schema during automated testing.
+
+We are not using Committee to validate incoming requests, as we handle all parameter validation and error formatting in controllers to keep that logic fully under our control.
```
kevindew, line 41: Should we consider using it? I could imagine there are some annoyances/pain points but I did wonder if we end up incompletely verifying our specification file by not using it as we could be only verifying the output and not the input.

docs/adr/0007-serialisation-and-schema-validation-for-govuk-chat-api.md
```diff
@@ -0,0 +1,57 @@
...
+### Serialisation
+
+We have chosen to use [Blueprinter](https://github.com/procore/blueprinter) for serializing objects into JSON. It provides a clean and lightweight way to define how data is exposed through the API.
+
+We decided against
+
+- **active_model_serializers**, as there are issues with maintenance and it has a history of stability issues. See:
+  - https://github.com/rails-api/active_model_serializers/issues
+  - https://github.com/rails-api/active_model_serializers/issues/2396
+- **jsonapi-serializer**, The team has had negative experiences with it.
```
kevindew, line 27: We should probably add some more information to this one since it's a bit ambiguous. 

```suggestion
- **jsonapi-serializer**, We felt that our returned data models would be harder to work with were they to follow the requirements of the JSON API specification and thus provide a reduced user experience.
```

kevindew (commented):
Looks great, thanks for taking the time to write this up. I've put a couple of suggestions in.

docs/adr/0007-serialisation-and-schema-validation-for-govuk-chat-api.md
davidgisbey: So this is an interesting one. I did consider whether we should use it initially but decided we would probably prefer our own custom error messaging rather than the generic error message that it provides `missing required: created at` etc.

Interestingly there is the ability to customise error messages [here](https://github.com/interagent/committee?tab=readme-ov-file#validation-errors), but only when raise is set to false.

I'm def not opposed to using it though.

davidgisbey: I've spiked this out in this [commit](https://github.com/alphagov/govuk-chat/pull/175/commits/db6060886ae6674ad3ad2310a9e4ee36a3733eec) and it's promising. It returns a 400 and a  useful error message.

The example of the default it provides in the docs is:

```
{
  "id":"invalid_response",
  "message":"Missing keys in response: archived_at, buildpack_provided_description, created_at, git_url, id, maintenance, name, owner:email, owner:id, region:id, region:name, released_at, repo_size, slug_size, stack:id, stack:name, updated_at, web_url."
}
```

I had a quick go at [building custom validation errors](https://github.com/interagent/committee?tab=readme-ov-file#validation-errors) and using them in place of the default. It was pretty straightforward. You just have to add the class that inherits from `Committee::ValidationError` and use pass it in as an option in the initialiser.

If we want we could use them to strip the id from the JSON so it is more similar to the structure we send back for 422s etc. Wdyt?

i.e.

```
{
  "message":"Missing keys in response: archived_at, buildpack_provided_description, created_at, git_url, id, maintenance, name, owner:email, owner:id, region:id, region:name, released_at, repo_size, slug_size, stack:id, stack:name, updated_at, web_url."
}
```

kevindew: Unfortunately the link to a commit isn't working for me but that does sound really promising and yeah looks like the option with just a message would be same JSON structure as our generic error which would be great.

Does look like it'll handle a bunch of cases: https://github.com/interagent/committee/blob/494146486074f7fb734907f3b737ad1a179efd6c/lib/committee/middleware/request_validation.rb#L12-L30 so I'd imagine we could hit some surprises/annoyances but lets try it out and see how it goes.

davidgisbey: oh sorry it's [this](https://github.com/alphagov/govuk-chat/pull/175/commits/9f6dab5e88adba6f39432b4511b296b1d14a052b) i'd pushed a change. Cool i'll add a commit for it 👍

davidgisbey: provided we're happy with the approach on that PR i'll update this when i'm back in from moving to reflect the new implementation

davidgisbey:
I've pushed up that we're going to validate requests at runtime too

kevindew (approved):
Nice one

PR #170, opened 2025-04-16
Implement initial Chat API endpoint GET answer

davidgisbey:
## Description

This PR introduces the initial implementation of the `GET /answer` endpoint for the GOV.UK Chat API. It: 

- Adds the OpenAI specification based on Kev's [Gist](https://gist.github.com/kevindew/9ee643132182ab4dc927c4c8ca46d8eb)
- Adds `blueprinter` as a dependency and adds the required blueprints for the endpoint (answer, answer source and generic error)
- Adds the endpoint (controller, route etc.)
- Adds `committee` and configured it to validate API responses at runtime
- Adds an `api-user` permission to the AdminUser model and auths the API endpoints based on the presence of the permission

For now you can access the API endpoints locally by adding the permission to the first admin user in your local DB. It's been added to the seeds though so when the db is reset it will work as expected.

A few things to note:

- I know we will need to rename the `AdminUser` model since we will have people who auth via a token to use the API who are not admins (currently when they use the token for the first time they will be persisted to the DB as an AdminUser)

I've carded up doing this as a separate bit of work [here](https://trello.com/c/zsmJJZfa/2420-chat-api-update-adminuser-to-signonuser), but am trying to get the rest of the cards for this unblocked first. 

- I've tried as much as possible to mirror our implementation for Chat where possible. Are couple of examples are:
  - using a single `ConversationsController` with shared functionality rather than `AnswerContoller#show` 
  - consistent error messaging (using `Conversation not found` rather than using `error.message` within the rescue

Im going to leave a couple of comments on bit's i'm less sure on and would appreciate feedback on.

## Trello card

https://trello.com/c/fTlGZAtH/2421-chat-api-add-initial-api-endpoint-get-answer

## Useful resources 

- We’re using the OpenAPI spec defined by Kev in this [Gist](https://gist.github.com/kevindew/9ee643132182ab4dc927c4c8ca46d8eb)

- We’ve decided on using [Blueprinter](https://github.com/procore-oss/blueprinter/blob/main/README.md) for serialisation. It was spiked out in this [PR](https://github.com/alphagov/govuk-chat/pull/153/files)

- We’re using [Committee](https://github.com/interagent/committee) to ensure we keep our responses in sync with the OpenAPI specification

- This [ADR](https://github.com/alphagov/govuk-chat/pull/167) discusses why we’re using them

### DIFF
--- a/app/blueprints/answer_blueprint.rb
+++ b/app/blueprints/answer_blueprint.rb
@@ -1,0 +1,17 @@
+class AnswerBlueprint < Blueprinter::Base
+  identifier :id
+
+  field :message
+  field :created_at do |answer, _options|
+    answer.created_at.iso8601
+  end
+  field :useful, if: ->(_field_name, answer, _options) { answer.feedback.present? } do |answer, _options|
+    answer.feedback.useful
+  end
+
+  field :sources, if: ->(_field_name, answer, _options) { answer.sources.used.present? } do |answer, _options|
+    answer.group_used_answer_sources_by_base_path.map do |source|
+      { url: source[:href], title: source[:title] }
+    end
+  end
+end
--- a/app/controllers/api/v0/conversations_controller.rb
+++ b/app/controllers/api/v0/conversations_controller.rb
@@ -1,0 +1,17 @@
+class Api::V0::ConversationsController < ApplicationController
+  before_action :find_conversation
+  before_action :find_question
+
+  def answer
+    conversation = Conversation.includes(questions: { answer: %i[sources feedback] })
+                               .find(params[:conversation_id])
+    question = conversation.questions.find(params[:question_id])
+    answer = question.answer
+
+    if answer.present?
+      render json: AnswerBlueprint.render(answer), status: :ok
+    else
+      render json: {}, status: :accepted
+    end
+  end
+end
--- a/app/controllers/api/v0/conversations_controller.rb
+++ b/app/controllers/api/v0/conversations_controller.rb
@@ -1,6 +1,5 @@
 class Api::V0::ConversationsController < ApplicationController
-  before_action :find_conversation
-  before_action :find_question
+  before_action { authorise_user!(SignonUser::Permissions::CONVERSATION_API) }
 
   def answer
     conversation = Conversation.includes(questions: { answer: %i[sources feedback] })
--- a/config/initializers/committee.rb
+++ b/config/initializers/committee.rb
@@ -1,0 +1,7 @@
+require "committee"
+
+Rails.application.config.middleware.use Committee::Middleware::ResponseValidation,
+                                        schema_path: "docs/api_openapi_specification.yml",
+                                        prefix: "/api/v0",
+                                        strict_reference_validation: true,
+                                        raise: true
--- a/spec/requests/api/v0/conversations_spec.rb
+++ b/spec/requests/api/v0/conversations_spec.rb
@@ -20,6 +20,19 @@ RSpec.describe "Api::V0::ConversationsController" do
     end
   end
 
+  describe "middleware ensures adherance to the OpenAPI specification" do
+    context "when the response returned does not conform to the OpenAPI specification" do
+      it "raises an error and returns the invalid params in the error message" do
+        create(:answer, question:)
+        allow(AnswerBlueprint).to receive(:render).and_return({}.to_json)
+
+        get api_v0_answer_question_path(conversation, question)
+        expect(response).to have_http_status(:internal_server_error)
+        expect(response.body).to include("Sorry, there is a problem with GOV.UK Chat")
+      end
+    end
+  end
+
   describe "GET :answer" do
     it_behaves_like "responds with forbidden if user doesn't have conversation-api permission",
                     :api_v0_answer_question_path,
--- a/spec/requests/api/v0/conversations_spec.rb
+++ b/spec/requests/api/v0/conversations_spec.rb
@@ -2,7 +2,31 @@ RSpec.describe "Api::V0::ConversationsController" do
   let(:conversation) { create(:conversation) }
   let(:question) { create(:question, conversation:) }
 
+  before do
+    login_as(create(:signon_user, :conversation_api))
+  end
+
+  shared_examples "responds with forbidden if user doesn't have conversation-api permission" do |path, method|
+    let(:route_params) { {} }
+    let(:params) { {} }
+
+    describe "responds with forbidden if user doesn't have conversation-api permission" do
+      it "returns with 403 for #{path}" do
+        login_as(create(:signon_user))
+        process(method.to_sym, public_send(path.to_sym, *route_params), params:, as: :json)
+
+        expect(response).to have_http_status(:forbidden)
+      end
+    end
+  end
+
   describe "GET :answer" do
+    it_behaves_like "responds with forbidden if user doesn't have conversation-api permission",
+                    :api_v0_answer_question_path,
+                    :get do
+      let(:route_params) { [create(:conversation), create(:question)] }
+    end
+
     context "when an answer has been generated for the question" do
       let!(:answer) { create(:answer, question:) }
 
--- a/spec/requests/api/v0/conversations_spec.rb
+++ b/spec/requests/api/v0/conversations_spec.rb
@@ -1,0 +1,44 @@
+RSpec.describe "Api::V0::ConversationsController" do
+  let(:conversation) { create(:conversation) }
+  let(:question) { create(:question, conversation:) }
+
+  describe "GET :answer" do
+    context "when an answer has been generated for the question" do
+      let!(:answer) { create(:answer, question:) }
+
+      it "returns a success status" do
+        get api_v0_answer_question_path(conversation, question)
+        expect(response).to have_http_status(:ok)
+      end
+
+      it "returns the expected JSON" do
+        get api_v0_answer_question_path(conversation, question)
+
+        eager_loaded_answer = Answer.includes(:sources, :feedback).find(answer.id)
+        expected_response = AnswerBlueprint.render_as_json(eager_loaded_answer)
+        expect(JSON.parse(response.body)).to eq(expected_response)
+      end
+
+      it "returns the correct JSON for answer sources" do
+        source = create(:answer_source, answer:)
+
+        get api_v0_answer_question_path(conversation, question)
+
+        expect(JSON.parse(response.body)["sources"])
+          .to eq([{ url: source.url, title: "#{source.title}: #{source.heading}" }.as_json])
+      end
+    end
+
+    context "when an answer has not been generated for the question" do
+      it "returns an accepted status" do
+        get api_v0_answer_question_path(conversation, question)
+        expect(response).to have_http_status(:accepted)
+      end
+
+      it "returns an empty JSON response" do
+        get api_v0_answer_question_path(conversation, question)
+        expect(JSON.parse(response.body)).to eq({})
+      end
+    end
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ ruby "~> #{File.read(File.join(__dir__, '.ruby-version')).strip}"
 gem "rails", "8.0.2"
 
 gem "aws-sdk-bedrockruntime"
+gem "blueprinter"
 gem "bootsnap"
 gem "chartkick"
 gem "csv"
--- a/Gemfile
+++ b/Gemfile
@@ -54,6 +54,7 @@ group :test do
 end
 
 group :development, :test do
+  gem "committee"
   gem "dotenv"
   gem "erb_lint", require: false
   gem "factory_bot_rails"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -110,6 +110,7 @@ GEM
       parser (>= 2.4)
       smart_properties
     bigdecimal (3.1.9)
+    blueprinter (1.1.2)
     bootsnap (1.18.4)
       msgpack (~> 1.2)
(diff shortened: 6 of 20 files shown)


app/blueprints/answer_blueprint.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  field :message
+  field :created_at do |answer, _options|
+    answer.created_at.iso8601
+  end
+  field :useful, if: ->(_field_name, answer, _options) { answer.feedback.present? } do |answer, _options|
+    answer.feedback.useful
+  end
+  association :sources, blueprint: AnswerSourceBlueprint,
+                        if: ->(_field_name, answer, _options) { answer.sources.present? }
```
davidgisbey, line 12: i just went back and looked at the spec [here](https://gist.github.com/kevindew/9ee643132182ab4dc927c4c8ca46d8eb#file-chat-openapi-yaml-L315-L317). I can see Kev has written 

```
URL of the page on GOV.UK that was used to generate answer with potentially a fragment to appropriate section
```

I'd assumed that we would give them the full url and they'd decide what to do with it, but since that's not the case i think we probably need to: 

1.  update this to be a field rather than an association 
2. extract the logic that handles [building the source urls ](https://github.com/alphagov/govuk-chat/blob/main/app/helpers/answers_helper.rb#L13-L28) for an answer to the model
3. use that in both places 

Does that sound about right?

app/controllers/api/v0/conversations_controller.rb
```diff
@@ -14,6 +15,10 @@ def answer
 
 private
 
+  def authorise_user
+    authorise_user!("api-user")
```
davidgisbey, line 19: `api-user` seems a bit over generic to me as we may add additional apis down the line i.e. this may become the `ConversationsApi`. I'm struggling to come up with a better name though since for now we just have the one API.

Does anyone have thoughts on a better name or do you think it's fine as is for now and we would just update the permission name down the line if required?

config/initializers/committee.rb
```diff
@@ -0,0 +1,7 @@
+require "committee"
+
+Rails.application.config.middleware.use Committee::Middleware::ResponseValidation,
+                                        schema_path: "spec/support/api/govuk_chat_api_specification.yaml",
+                                        prefix: "/api/v0",
+                                        strict_reference_validation: true,
+                                        raise: true
```
davidgisbey, line 7: I thought i'd just lay out my thoughts on why i've done this and see if anyone disagrees/has any input.

This middleware runs on responses at runtime and during tests. This means that if any of our API request specs don't adhere to the schema the test will fail with 

image

etc.

Which unless i'm missing something really just means that we need a test that checks we get a failure in a test when the response doesn't adhere to the schema for each endpoint in the specification.

The other point i'd like some feedback on is the `raise: true`.

If don't set this it defaults to false. This means that the response the client get's would contain 

`#/components/schemas/Answer/properties/created_at Value: "2025-04-16 11:51:34 +0100" is not conformant with date-time format` but we will not be informed via Sentry. It'll just log it and silently fail.

Since users of the API cannot actually fix anything, i figured it'd be better to raise it so we get pinged on Slack.

Other options are: 

- don't run the middleware and rely on testing each endpoint like [this](https://github.com/alphagov/govuk-chat/pull/163/commits/bf604652441e6e37c7e4dc70016d866a8f40e22c#diff-822488d4d54203be47042a30ff6f804a348624128bef23500021ec2704fa4455R27-R30) 
- setting raise to false and relying on API consumers to inform us when there are issues

To me it makes more sense for things to fail loudly and quickly if there's an issue, but it would mean if something was shipped with a minor issue it might end up causing an incident.

For example, say 1 minor fairly unimportant field was missing from the schema it would fail rather than delivering the payload with the missing field.

Realistically though due to the middleware that should be caught prior to merge provided we've got good test coverage of the combinations of responses we can deliver for the endpoint.

app/models/admin_user.rb
```diff
@@ -3,6 +3,7 @@ class AdminUser < ApplicationRecord
 
   module Permissions
     ADMIN_AREA = "admin-area".freeze
```
davidgisbey, line 5: I've noted this in the PR description, but we're going to need to rename this model since not all users persisted in our DB will be admin users:

- some users will arrive from Signon with a full account and realistically will be admin users
- the app with auth via bearer token and will not be an admin user. It'll just need to consume the API

davidgisbey:
well that wasn't much use 😂

app/controllers/api/v0/conversations_controller.rb
chaecramb: Yeah, see what you mean. I suppose it could be something like `conversation-api-reader` down the line, but being that specific now seems preemptive.

It seems reasonable to keep it generic just now, and update when needed.

config/initializers/committee.rb
chaecramb: I'd agree, being stricter with `raise: true` seems the better default to me, to ensure adherence to the OpenAPI spec. 

The only downside seems to be that it requires thorough test coverage to avoid accidental failures, but then that's not really much of a downside, since we should do it anyway. Whereas `raise: false` would give us less visibility of problems.

davidgisbey: Thanks Chae. From what i remember working on the Publishing team when there's an issue with schema adherence it raises there too so i suspect it's fine

app/controllers/api/v0/conversations_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+    @conversation = Conversation.includes(questions: { answer: %i[sources feedback] })
+                                .find(params[:conversation_id])
+  rescue ActiveRecord::RecordNotFound
+    render json: GenericErrorBlueprint.render(message: "Conversation not found"), status: :not_found
+  end
+
+  def find_question
+    @question = @conversation.questions.find(params[:question_id])
+  rescue ActiveRecord::RecordNotFound
+    render json: GenericErrorBlueprint.render(message: "Question not found"), status: :not_found
```
chaecramb, line 32: Not sure is this actually matters, but the API spec has a description "Conversation or question do not exist".

Differentiating between "Question not found" and "Conversation not found" as you do, seems more helpful, but is this something that should be updated in the OpenAPI spec? I don't have experience with it, so not sure if it's supposed to reflect all messages exactly.

chaecramb (approved):
Looks good. Just a couple of comments that are more queries.

app/controllers/api/v0/conversations_controller.rb
davidgisbey: So in an OpenAPI spec the description just describes the so for example a 422 returned on the conversation POST endpoint has the description of 

```
```

davidgisbey:
great shout on `API_USER` 👍

app/blueprints/answer_blueprint.rb
```diff
@@ -0,0 +1,17 @@
...
+  field :created_at do |answer, _options|
+    answer.created_at.iso8601
+  end
+  field :useful, if: ->(_field_name, answer, _options) { answer.feedback.present? } do |answer, _options|
+    answer.feedback.useful
+  end
+
+  field :sources, if: ->(_field_name, answer, _options) { answer.sources.used.present? } do |answer, _options|
+    answer.group_used_answer_sources_by_base_path.map do |source|
+      source.transform_keys { |key| key == :href ? :url : key }
```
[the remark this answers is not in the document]
davidgisbey: Great shout this far simpler 👍

app/controllers/api/v0/conversations_controller.rb
davidgisbey: yea great shout!

kevindew (commented):
Looks great, nice job. I've got a few actionable comments there and I'll have a quick research about the error stuff, maybe we should have a quick chat about it.

spec/support/api/govuk_chat_api_specification.yaml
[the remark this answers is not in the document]
davidgisbey: yea sorry i did have to move it over when i was playing around with Rswag::UI then forgot to do it on my main branch  and cherry-picked the commit. Will sort that

config/initializers/committee.rb
davidgisbey: yea it returns a 500 with a v similar message to PubAPI 
image

app/controllers/api/v0/conversations_controller.rb
kevindew: So I've had a little look and think about it. 

I think the approach I'd suggest doing is that we have a check in ErrorsController for it. I imagine the most Railsy approach would be to do the standard responds_to 

```
  def not_found
    respond_to do |format|
      format.json do
        render json: GenericErrorBlueprint.render(message: "Resource not found"),
               status: :not_found
      end
      format.any { render status: :not_found, formats: :html }
    end
  end
```

and do the same for each method in ErrorsController.

If we need to do anything more complex (like different ones for different API versions) or if we hit compatibility annoyances (since we return JSON in the app too) we could potentially query `request.original_fullpath.start_with?("/api")` but I think `respond_to` might be a more idiomatic Rails idiomatic place to start from.

davidgisbey: cool yea that'll def work for now 👍 I'll make the change

app/controllers/api/v0/conversations_controller.rb
```diff
@@ -0,0 +1,34 @@
+class Api::V0::ConversationsController < ApplicationController
+  before_action :authorise_user
```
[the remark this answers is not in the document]
davidgisbey: So this ones interesting because i'm not sure we're going to end up with more than 1 controller for our initial implementation of the API.

I've mirrored the implementation of our other chat controller where possible. i.e gone for `conversations#answer` rather than `answers#show` so we can share functionality between actions (finding convos, questions, answers etc.). 

I do expect that down the line we will need a base controller though so happy to implement one now if you'd prefer it 👍 Lmk what you think 

Btw this is now `before_action { authorise_user!(SignonUser::Permissions::CONVERSATION_API) }`

davidgisbey:
Thanks for the review @kevindew. I've addressed the comments except this [one](https://github.com/alphagov/govuk-chat/pull/170#discussion_r2054394008) as i've mentioned i think we will only be using one controller in the short to mid term, but v happy to implement the base controller if you'd like.

app/controllers/api/v0/conversations_controller.rb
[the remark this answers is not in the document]
davidgisbey: we've agreed to leave it as is for now with the expectation that we will add a base controller if/when a new controllers added.

davidgisbey:
@kevindew just an fyi i made one change after we spoke based on our convo. I moved the test that requests the response is valid to it's [own describe block](https://github.com/alphagov/govuk-chat/pull/170/commits/644299435ff6fd7d6439ff51d91c8bf1e51cee89#diff-822488d4d54203be47042a30ff6f804a348624128bef23500021ec2704fa4455R9-R24)

kevindew (commented):
Looks great, I think there's a couple of things to move around and I'd suggest we pull out the ErrorController stuff into a separate PR to get this one through quicker.

app/controllers/errors_controller.rb
```diff
@@ -14,7 +14,13 @@ def forbidden
...
 
   def not_found
-    render status: :not_found, formats: :html
+    respond_to do |format|
+      format.json do
+        render json: GenericErrorBlueprint.render(message: "Resource not found"),
+               status: :not_found
+      end
+      format.any { render status: :not_found, formats: :html }
+    end
```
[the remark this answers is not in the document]
davidgisbey: okay i've removed for now. I'll add it in a follow up PR 👍

spec/requests/api/v0/conversations_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    context "when the response returned does not conform to the OpenAPI specification" do
+      it "raises an error and returns the invalid params in the error message" do
+        answer = create(:answer, question:)
+        eager_loaded_answer = Answer.includes(:sources, :feedback).find(answer.id)
+        answer_blueprint = AnswerBlueprint.render_as_hash(eager_loaded_answer)
+        answer_blueprint.delete(:created_at)
+        allow(AnswerBlueprint).to receive(:render).and_return(answer_blueprint)
+
+        expect { get api_v0_answer_question_path(conversation, question) }
+          .to raise_error(Committee::InvalidResponse) do |error|
```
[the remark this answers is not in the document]
davidgisbey: cool have done. Since it's now returning our 500 from the errors controller but the follow up PR will sort that

spec/requests/api/v0/conversations_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+        expect(JSON.parse(response.body)).to eq({})
+      end
+    end
+
+    context "when the user does not have the correct permissions to use the API" do
+      it "returns a forbidden status" do
+        login_as(create(:signon_user))
+        get api_v0_answer_question_path(conversation, question)
+        expect(response).to have_http_status(:forbidden)
+      end
```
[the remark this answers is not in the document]
davidgisbey: cool i've added one in the follow up PR but i can cherry-pick that into this commit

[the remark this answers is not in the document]
kevindew: Feel free to sort it in that PR

kevindew (approved):
Nice one 🚀

spec/requests/api/v0/conversations_spec.rb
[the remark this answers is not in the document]
davidgisbey: just sorted it here

PR #172, opened 2025-04-17
Rename AdminUser to SignonUser

davidgisbey:
## Description

We're going to have multiple types of users arriving from Signon. Admin users who have full Signon accounts and API users who have a bearer token which they'll use when hitting our API endpoints.

AdminUser doesn't reflect this new use case since they won't be admin and shouldn't have admin permissions. I've proposed we update it to SignonUser rather than just User because we've got a bunch of user types still knocking about (WaitingListUser, EarlyAccessUser etc.)

Since we're not live i've not worried about renaming the columns and doing the usual:

- add new columns
- update code to write to both
- backfill
- switch over to new column 
- delete old table/columns

Since we shouldn't experience any issues as we're extremely unlikely to creating new AdminUsers during the deploy as no-one is using the application in the near future.

## Trello card

https://trello.com/c/zsmJJZfa/2420-chat-api-update-adminuser-to-signonuser

### DIFF
--- a/app/controllers/admin/base_controller.rb
+++ b/app/controllers/admin/base_controller.rb
@@ -1,5 +1,5 @@
 class Admin::BaseController < ApplicationController
   layout "admin"
-  before_action { authorise_user!(AdminUser::Permissions::ADMIN_AREA) }
-  before_action { Current.admin_user = current_user }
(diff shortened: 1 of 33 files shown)


kevindew (approved):
Looks good to me 👍

PR #176, opened 2025-04-23
Fix failing SignonUser migration

davidgisbey:
## Description 

The columns were meant to be renamed in the migration, but both the current and new column name were the same which caused the migration to fail.


kevindew (approved):
Nice one

PR #178, opened 2025-04-23
Update ErrorsController to handle JSON requests

davidgisbey:
We don't want to render back our HTML error pages when the requests content type is JSON.

This updates the controller to respond with a JSON error message when the request is made with a JSON content type.

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -6,26 +6,62 @@ class ErrorsController < BaseController
   after_action { response.headers["No-Fallback"] = "true" }
 
   def bad_request
-    render status: :bad_request, formats: :html
+    respond_to do |format|
+      format.json do
+        render json: GenericErrorBlueprint.render(message: "Bad request"),
+               status: :bad_request
+      end
+      format.any { render status: :bad_request, formats: :html }
+    end
   end
 
   def forbidden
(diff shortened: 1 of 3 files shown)


app/controllers/errors_controller.rb
```diff
@@ -6,26 +6,62 @@ class ErrorsController < BaseController
...
+      end
+      format.any { render status: :forbidden, formats: :html }
+    end
   end
 
   def not_found
-    render status: :not_found, formats: :html
+    respond_to do |format|
+      format.json do
+        render json: GenericErrorBlueprint.render(message: "Resource not found"),
```
davidgisbey, line 31: this is the only error message which i've made different to the page title used. We may want to use different texts, but i'm largely thinking them being generic and reusing the page titles is fine (` "Sorry, there is a problem"` etc.)

kevindew: I think given we expect the API to be machine consumed it'd be better if these were less user friendly and more just described the status code (basically cutting out the need to check the HTTP status).

So I'd suggest we have them basically match the status code and be: Bad request, Forbidden, Not found etc

davidgisbey: okay cool i'll sort that now

kevindew (approved):
Nice one

PR #183, opened 2025-04-24
Allow docs/api_openapi_specification.yml in Docker

kevindew:
The openapi specification is a runtime dependency so needs to be included in a docker build. I've taken the step of just allowing that one file so we don't put all the other data in, this might prove foolish as we'll hit another annoying issue if we add another openapi file or rename this one. Let's see.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,7 +7,8 @@ Dockerfile
 Procfile
 README.md
 coverage
-docs
+docs/**
+!docs/api_openapi_specification.yml
 features
 log
 node_modules


jackbot (approved):
LGTM, but it does feel a bit weird to me that the spec file is in the `docs` directory. I can see the rationale, but it feels to me like `docs` is for human-readable documentation which this _kind of_ is, but also kind of isn't as it's very much an app dependency too.

kevindew:
Yup I know what you mean, I’m not sure where seems better though.

config feels wrong for others to use it as docs but right for us as a runtime config dependency
Root could be an option but feels a bit of a cop out
Could follow search-api-v2 and have a docs/openapi but that’s still in docs
Could have a root openapi folder but that feels a bit overkill

PR #184, opened 2025-04-24
Add GET /conversations/{id} endpoint to API

jackbot:
https://trello.com/c/cdHh2shY/2422

Adds the blueprints and API endpoint for retrieving a conversation by its ID.

### DIFF
--- a/app/blueprints/conversation_blueprint.rb
+++ b/app/blueprints/conversation_blueprint.rb
@@ -1,0 +1,22 @@
+class ConversationBlueprint < Blueprinter::Base
+  identifier :id
+
+  field :created_at, datetime_format: ->(datetime) { datetime.iso8601 }
+
+  association(
+    :answered_questions,
+    blueprint: QuestionBlueprint,
+    view: :answered,
+  ) do |_conversation, options|
+    options[:answered_questions] || []
+  end
+
+  association(
+    :pending_question,
+    blueprint: QuestionBlueprint,
+    view: :pending,
+    if: ->(_field_name, _conversation, options) { options[:pending_question].present? },
+  ) do |_conversation, options|
+    options[:pending_question]
+  end
+end
--- a/app/controllers/api/v0/conversations_controller.rb
+++ b/app/controllers/api/v0/conversations_controller.rb
@@ -1,6 +1,18 @@
 class Api::V0::ConversationsController < ApplicationController
   before_action { authorise_user!(SignonUser::Permissions::CONVERSATION_API) }
 
+  def show
+    conversation = Conversation.includes(questions: { answer: %i[sources feedback] })
+                               .find(params[:conversation_id])
+    answered_questions = conversation.questions.joins(:answer)
+    pending_question = conversation.questions.unanswered.last
+
+    render(
+      json: ConversationBlueprint.render(conversation, answered_questions:, pending_question:),
+      status: :ok,
+    )
+  end
+
   def answer
     conversation = Conversation.includes(questions: { answer: %i[sources feedback] })
                                .find(params[:conversation_id])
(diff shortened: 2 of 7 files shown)


app/blueprints/conversation_blueprint.rb
```diff
@@ -0,0 +1,13 @@
...
+  identifier :id
+
+  field :created_at, datetime_format: ->(datetime) { datetime.iso8601 }
+
+  association :pending_question, blueprint: QuestionBlueprint, view: :pending do |conversation, _options|
+    conversation.questions.unanswered.last
+  end
+
+  association :answered_questions, blueprint: QuestionBlueprint, view: :answered do |conversation, _options|
+    conversation.questions.joins(:answer).order(created_at: :desc)
```
davidgisbey, line 11: This is how i originally implemented this too. Kev was of the opinion that custom associations should be [injected via options on init](https://github.com/alphagov/govuk-chat/pull/153#discussion_r2037065402) rather than within the blueprint itself. Wdyt?

davidgisbey (commented):
looks great. Just left a couple of minor comments for discussion

spec/requests/api/v0/conversations_spec.rb
```diff
@@ -42,8 +29,15 @@
       let(:route_params) { [create(:conversation)] }
     end
 
+    it "raises an error if the response does not adhere to the OpenAPI spec" do
```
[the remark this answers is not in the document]
jackbot: Yeah I think that makes sense. I've reverted it to just have a single test.

spec/blueprints/conversation_blueprint_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+          conversation,
+          answered_questions: answered_questions,
+          pending_question: pending_question,
+        )
+
+        expect(output_json).to eq(expected_json)
+      end
+    end
+
+    context "with no answered questions passed in" do
```
davidgisbey, line 55: Sorry this might have been me if you grabbed it from the draft PR but i think this is covered by  https://github.com/alphagov/govuk-chat/pull/184/commits/09edc9441fc502ae22e2f42f3e378b16edb36ac3#diff-2bffad8ec99d11d9143ec51d3fc9d71da636c4c48075950a5a3bd0e68edfaba6R5-R24

If you feel like it's worth explicity pointing out that it renders an empty array for answers (i'm not sure it does) then we could add an additional test there instead

app/controllers/api/v0/conversations_controller.rb
```diff
@@ -1,6 +1,16 @@
 class Api::V0::ConversationsController < ApplicationController
   before_action { authorise_user!(SignonUser::Permissions::CONVERSATION_API) }
 
+  def show
+    conversation = Conversation.includes(questions: { answer: %i[sources feedback] })
+                               .find(params[:conversation_id])
+    answered_questions = conversation.questions.joins(:answer)
+    pending_question = conversation.questions.unanswered.last
+
+    response = ConversationBlueprint.render(conversation, answered_questions:, pending_question:)
```
davidgisbey: not a biggie but maybe use `json =  ConversationBlueprint.render(conversation, answered_questions:, pending_question:)` rather than response since it''ll overwrite the response object.

davidgisbey (commented):
awesome work. Couple of minor comments but it's basically good to go

davidgisbey (approved):
comments are super nitty and optional so i'll leave it up to you to make the changes if you think they're worth it

app/controllers/api/v0/conversations_controller.rb
jackbot: Great spot. I've just removed the variable entirely as it's not really necessary.

PR #194, opened 2025-04-29
Add POST /api/v0/conversation endpoint 

Sunny-xyz:
Trello: https://trello.com/c/INvwAr2v/2423-chat-api-add-post-conversation-endpoint

We’re building an API endpoint that the app team will use to integrate with GOV.UK Chat; this PR adds a POST /api/v0/conversation action in Api::V0::ConversationsController so that when a user asks their first question, we persist a new Conversation and Question in our database and return a JSON payload (via the QuestionBlueprint pending view). All validation is delegated to the existing Form::CreateQuestion to avoid duplicating logic. I introduced a question_params method, and wired up the new route in routes.rb. New request specs cover question validity, ensuring the endpoint behaves exactly as expected.

### DIFF
--- a/app/controllers/api/v0/conversations_controller.rb
+++ b/app/controllers/api/v0/conversations_controller.rb
@@ -2,6 +2,19 @@ class Api::V0::ConversationsController < ApplicationController
   before_action { authorise_user!(SignonUser::Permissions::CONVERSATION_API) }
   before_action :find_conversation, only: %i[show answer answer_feedback]
 
+  def create
+    conversation = Conversation.new
+
+    form = Form::CreateQuestion.new(conversation:, user_question: question_params[:user_question])
+
+    if form.valid?
+      question = form.submit
+      render json: QuestionBlueprint.render(question, view: :pending), status: :created
+    else
+      render json: ValidationErrorBlueprint.render(message: "Could not create question", errors: form.errors.messages), status: :unprocessable_entity
+    end
+  end
+
   def show
     answered_questions = @conversation.questions.joins(:answer)
     pending_question = @conversation.questions.unanswered.last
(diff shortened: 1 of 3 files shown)


spec/requests/api/v0/conversations_spec.rb
```diff
@@ -89,6 +96,51 @@
...
+        end
+
+        it "returns the correct JSON in the body" do
+          post api_v0_create_conversation_path, params: payload, as: :json
+
+          expect(JSON.parse(response.body))
+            .to eq(
+              {
+                "message" => "Unprocessable entity",
+                "errors" => { "user_question" => [Form::CreateQuestion::USER_QUESTION_PRESENCE_ERROR_MESSAGE] },
```
davidgisbey, line 136: nice one using the constant here 👍

app/controllers/api/v0/conversations_controller.rb
```diff
@@ -2,6 +2,19 @@ class Api::V0::ConversationsController < ApplicationController
   before_action { authorise_user!(SignonUser::Permissions::CONVERSATION_API) }
   before_action :find_conversation, only: %i[show answer answer_feedback]
 
+  def create
+    conversation = Conversation.new
+
+    form = Form::CreateQuestion.new(conversation: conversation, user_question: question_params[:user_question])
```
davidgisbey, line 8: i'm surprised rubocop isn't complaining about this, but it used to (and maybe should be)

With hashes if you have a variable assigned you don't need to pass it in as the value. So this
```
form = Form::CreateQuestion.new(conversation: conversation, user_question: question_params[:user_question])
```
can be 

```
form = Form::CreateQuestion.new(conversation, user_question: question_params[:user_question])
```

davidgisbey: it must have been removed from govuk-rubocop as we've got 23 cases of it in our codebase so dw about it for now

davidgisbey (approved):
great work 👍

PR #195, opened 2025-04-30
Add PUT conversation endpoint

davidgisbey:
## Description

This adds the PUT conversation endpoint. This endpoint expects a user_question and returns:

- 201 if the params are valid. It creates a new question on the conversation
- 422 if the params are invalid
- 400 for requests that don't adhere to the schema
- 404 if the conversation cannot be found 

I've done a minor clean up in the CreateQuestion form so we're consistent in applying the use of constants for our error messages.

## Trello card

https://trello.com/c/wSIOGv6W/2424-chat-api-add-put-conversation-endpoint

### DIFF
--- a/app/models/form/create_question.rb
+++ b/app/models/form/create_question.rb
@@ -9,6 +9,9 @@ class Form::CreateQuestion
   USER_QUESTION_PRESENCE_ERROR_MESSAGE = "Ask a question. For example, 'how do I register for VAT?'".freeze
   USER_QUESTION_LENGTH_MAXIMUM = 300
   USER_QUESTION_LENGTH_ERROR_MESSAGE = "Question must be %{count} characters or less".freeze
+  USER_QUESTION_PII_ERROR_MESSAGE = "Personal data has been detected in your question. Please remove it and try asking again.".freeze
+  QUESTION_LIMIT_REACHED_MESSAGE = "You’ve reached the message limit for the GOV.UK Chat trial. You have no messages left.".freeze
+  UNANSWERED_QUESTION_ERROR_MESSAGE = "Previous question pending. Please wait for a response".freeze
 
   before_validation :sanitise_user_question
 
@@ -51,15 +54,13 @@ private
 
   def all_questions_answered?
     if conversation.questions.unanswered.exists?
-      errors.add(:base, "Previous question pending. Please wait for a response")
+      errors.add(:base, UNANSWERED_QUESTION_ERROR_MESSAGE)
     end
   end
 
   def no_pii_present?
     if PiiValidator.invalid?(user_question)
-      error_message = "Personal data has been detected in your question. Please remove it and try asking again."
-
-      errors.add(:user_question, error_message)
+      errors.add(:user_question, USER_QUESTION_PII_ERROR_MESSAGE)
     end
   end
 
@@ -67,6 +68,6 @@ private
     return if conversation.user.nil?
     return unless conversation.user.question_limit_reached?
 
-    errors.add(:base, "You’ve reached the message limit for the GOV.UK Chat trial. You have no messages left.")
+    errors.add(:base, QUESTION_LIMIT_REACHED_MESSAGE)
   end
 end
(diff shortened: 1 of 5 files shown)


spec/requests/api/v0/conversations_spec.rb
```diff
@@ -6,38 +6,52 @@
...
-    let(:route_params) { [SecureRandom.uuid, SecureRandom.uuid] }
-  end
-
-  it_behaves_like "responds with forbidden if user doesn't have conversation-api permission",
-                  :api_v0_answer_feedback_path,
-                  :post do
-    let(:route_params) { [SecureRandom.uuid, SecureRandom.uuid] }
-    let(:params) { { useful: true } }
-  end
+                  routes: [
```
jackbot, line 29: Great improvement!

app/models/form/create_question.rb
```diff
@@ -9,6 +9,9 @@ class Form::CreateQuestion
   USER_QUESTION_PRESENCE_ERROR_MESSAGE = "Ask a question. For example, 'how do I register for VAT?'".freeze
   USER_QUESTION_LENGTH_MAXIMUM = 300
   USER_QUESTION_LENGTH_ERROR_MESSAGE = "Question must be %{count} characters or less".freeze
+  USER_QUESTION_PII_ERROR_MESSAGE = "Personal data has been detected in your question. Please remove it and try asking again.".freeze
+  QUESTION_LIMIT_REACHED_MESSAGE = "You’ve reached the message limit for the GOV.UK Chat trial. You have no messages left.".freeze
```
jackbot, line 13: Wonder if we should remove the word "trial" here, as we're no longer in the public beta?

app/models/form/create_question.rb
```diff
@@ -9,6 +9,9 @@ class Form::CreateQuestion
   USER_QUESTION_PRESENCE_ERROR_MESSAGE = "Ask a question. For example, 'how do I register for VAT?'".freeze
   USER_QUESTION_LENGTH_MAXIMUM = 300
   USER_QUESTION_LENGTH_ERROR_MESSAGE = "Question must be %{count} characters or less".freeze
+  USER_QUESTION_PII_ERROR_MESSAGE = "Personal data has been detected in your question. Please remove it and try asking again.".freeze
```
jackbot, line 12: Are we keeping with the human readable error messages here then, even though it's being consumed via an API?

I'm wondering how these errors will work within the app. For stuff like server errors I expect the app will just check for a `500` status code and render their own error message to the user. But for stuff like this it feels like they'd actually want to either render this specific string in the app (or something similar), so the user can act on it?

I guess I'm wondering how we distinguish between a more generic error message and something that we'd actually want to show the user.

jackbot (commented):
Looks good, just a couple of questions about error messaging.

app/models/form/create_question.rb
davidgisbey: Yea good shout. I'm not sure whether we should be doing that as and when or whether it will be done as part of a content audit down the line. Happy to remove it though if you are 👍

app/models/form/create_question.rb
davidgisbey: Yea i was chatting to Kev about that and i think you're bang on. Generally they're going to just get a generic JSON error defined in the errors controller (for 500s etc.), but for 422s we're always going to send them specific error messages to render to the user and the field as the key so they can link the errors to the relevant input.

jackbot: Nice one, thanks for the explanation. All sounds good to me.

app/models/form/create_question.rb
jackbot: Yeah maybe we leave it until the audit. I guess the question limit number or logic might be tweaked too which would inform this error message so let's leave it as-is.

jackbot (approved):
Nice one 👍

davidgisbey:
thanks for the review. I'll let Sunny merge first so i don't hit him with even more merge conflicts 😂

PR #198, opened 2025-05-01
Use configured answer strategy in eval report generator

jackbot:
It'd be useful to be able to set the `ANSWER_STRATEGY` env var when
running the `evaluation:generate_report` Rake task to override the
provider, as we allow in multiple other places in the app.

Currently we build the question object manually which will use the
database default `answer_strategy`. This commit explicitly sets it to be
the Rails configuration value which uses the `ANSWER_STRATEGY` env var.


kevindew (approved):
Nice one

PR #199, opened 2025-05-01
Associate conversations created by api with the current signon user

davidgisbey:
## Description

We need to be able to associate conversations created by the API with the SignonUser (API user) that made the request. Down the line we suspect we will have multiple API users and want to be able to distinguish between conversations belonging to each. As well as between API users and web users.

This: 
- adds a signon_user foreign key to the conversations table
- creates the association in the models
- updates the create endpoint to associate the conversation with the current user (SignonUser)

## Out of scope 

- changes to the Admin UI

## Trello card

https://trello.com/c/3979ssqM/2447-chat-api-associate-conversations-created-by-api-with-user

### DIFF
--- a/app/controllers/api/v0/conversations_controller.rb
+++ b/app/controllers/api/v0/conversations_controller.rb
@@ -1,4 +1,5 @@
 class Api::V0::ConversationsController < ApplicationController
+  skip_before_action :verify_authenticity_token
   before_action { authorise_user!(SignonUser::Permissions::CONVERSATION_API) }
   before_action :find_conversation, only: %i[show update answer answer_feedback]
 
--- a/app/controllers/api/v0/conversations_controller.rb
+++ b/app/controllers/api/v0/conversations_controller.rb
@@ -69,6 +69,7 @@ private
   def find_conversation
     @conversation = Conversation
                     .includes(questions: { answer: %i[sources feedback] })
+                    .where(signon_user_id: current_user.id)
                     .find(params[:conversation_id])
   end
 
--- a/app/controllers/api/v0/conversations_controller.rb
+++ b/app/controllers/api/v0/conversations_controller.rb
@@ -3,7 +3,7 @@ class Api::V0::ConversationsController < ApplicationController
   before_action :find_conversation, only: %i[show update answer answer_feedback]
 
   def create
-    conversation = Conversation.new
+    conversation = Conversation.new(signon_user: current_user)
     form = Form::CreateQuestion.new(question_params.merge(conversation:))
 
     if form.valid?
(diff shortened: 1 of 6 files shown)


db/migrate/20250501150559_add_signon_user_foreign_key_to_conversation.rb
```diff
@@ -0,0 +1,5 @@
+class AddSignonUserForeignKeyToConversation < ActiveRecord::Migration[8.0]
+  def change
+    add_reference :conversations, :signon_user, foreign_key: { null: true, on_delete: :nullify }, type: :uuid
```
kevindew, line 3: on delete nullify seems a little fishy to me given we'd just lose history. To avoid boiling the ocean here I think going for `on_delete: restrict` as an interim measure would be prudent. 

I imagine a potential future option is that we don't have an actual foreign key and just retain the id if a signon user is deleted, or we can cope with it being null as we have additional metadata that tells us where the request came from.

app/controllers/api/v0/conversations_controller.rb
```diff
@@ -3,7 +3,7 @@ class Api::V0::ConversationsController < ApplicationController
   before_action :find_conversation, only: %i[show update answer answer_feedback]
 
   def create
-    conversation = Conversation.new
+    conversation = Conversation.new(signon_user: current_user)
```
kevindew, line 6: I imagine we also want this to be part of the where query for looking up a conversation. As, with the code currently, we'd allow any signon user to carry on the conversation or look up aspects of it.

kevindew (commented):
Looks good, just a couple of considerations that we should look at.

app/controllers/api/v0/conversations_controller.rb
davidgisbey: ah great catch 👍 Will sort

db/migrate/20250501150559_add_signon_user_foreign_key_to_conversation.rb
davidgisbey: Yea i was bouncing between which one to use but restrict makes sense to me. Realistically i can't see a reason the user would be deleted from the db and the convos wouldn't also be, so i'm not opposed to just retaining the ID. 

I'll update it to restrict for now though

davidgisbey:
@kevindew thanks for that. I've made those changes 👍

spec/requests/api/v0/conversations_spec.rb
```diff
@@ -104,6 +105,15 @@
 
       expect(response).to have_http_status(:not_found)
     end
+
+    it "returns a 404 if the conversation is not associated with the user" do
```
kevindew, line 109: Looks good - I imagine a potential semantic improvement could be changing this to be a 403, but it doesn't seem worth it now with the expectation of so few clients.

kevindew:
Yup good shout on the CSRF token - something for a potential api base controller

PR #205, opened 2025-05-06
Switch all Claude models to 3.7 Sonnet

jackbot:
https://trello.com/c/7gnhh5TG/2459-switch-claude-models-from-35-to-37

This is available on Bedrock now so we want to experiment with using it.

Rather than hardcoding the model ID all over the place, it's been
extracted to a constant in a `BedrockModels` module.


kevindew (approved):
Nice one

PR #207, opened 2025-05-06
Namespace OpenAI embedding class

jackbot:
Move TextToEmbedding class to OpenAI namespace and allow embedding provider to be configured.

This starts the work to allow us to use different embedding providers by namespacing the current provider and allowing configuration by an environment variable.

As it stands, this change does not affect the current functionality but sets the stage for future embedding providers.

The next steps will be:

1. Add a `Search::TextToEmbedding::Claude` class to make the Bedrock API call for some text to convert it to an embedding
2. Update `MessageQueue::ContentSynchroniser::IndexContentItem` to index GOV.UK content using both OpenAI and Claude embeddings and store both of the results in OpenSearch against the document
3. Extend the search results fetcher to use the Claude embedding provider when retrieving documents if the env var is set

### DIFF
--- a/lib/search/results_for_question.rb
+++ b/lib/search/results_for_question.rb
@@ -17,6 +17,7 @@ module Search
       search_start_time = Clock.monotonic_time
       results = ChunkedContentRepository.new.search_by_embedding(embedding, max_chunks:)
       metrics[:search_duration] = Clock.monotonic_time - search_start_time
+      metrics[:embedding_provider] = provider
 
       reranking_start_time = Clock.monotonic_time
       weighted_results = Search::ResultsForQuestion::Reranker.call(results)
--- a/lib/search/results_for_question.rb
+++ b/lib/search/results_for_question.rb
@@ -7,9 +7,11 @@ module Search
       max_results = Rails.configuration.search.thresholds.max_results
       max_chunks = Rails.configuration.search.thresholds.retrieved_from_index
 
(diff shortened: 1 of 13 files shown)


lib/search/results_for_question.rb
```diff
@@ -7,14 +7,26 @@ def self.call(question_message)
...
+
       metrics = {}
       embedding_start_time = Clock.monotonic_time
-      embedding = Search::TextToEmbedding.call(question_message)
+      embedding = case provider.to_sym
+                  when :openai
+                    Search::TextToEmbedding::OpenAI.call(question_message)
+                  else
+                    raise "Unknown embedding provider: #{provider}"
+                  end
```
kevindew, line 19: I wonder if rather than having this a case statement here (and not having one in [lib/message_queue/content_synchroniser/index_content_item.rb](https://github.com/alphagov/govuk-chat/pull/207/files#diff-fd6a3f5da0e3109ea12e83faf72d7b2e1819be532fa30739ec269203b1a16def)) we might instead want to abstract it a way so both can have it easy with an interface like:

Search::TextToEmbedding.call(provider, to_embed)

and have that check the provider and raise the exception?

kevindew (commented):
Looks great, just a suggestion that might make it a little cleaner.

davidgisbey:
looks great 👍

kevindew (approved):
Nice one

PR #208, opened 2025-05-06
Fix Claude 3.7 Sonnet model ID

jackbot:
We're using this model via Bedrock's cross region inference, meaning we
need to use the Inference profile ID rather than the generic model ID.


kevindew (approved):
Nice, has it run ok locally?

PR #209, opened 2025-05-06
Add ability to filter questions in Admin UI by source 

davidgisbey:
## Description

This PR does a few things:

- adds a new enum for conversation_source to the db
- adds a source column to the conversations table for so which defaults to :web
- updates the API controller to set the source value to :API when a conversation is created
- adds a filter to the questions index page which allows you to filter on source

image

## Trello card

https://trello.com/c/euKSVD56/2449-chat-api-add-admin-ui-functionality-for-api-users

### DIFF
--- a/app/controllers/api/v0/conversations_controller.rb
+++ b/app/controllers/api/v0/conversations_controller.rb
@@ -4,7 +4,7 @@ class Api::V0::ConversationsController < ApplicationController
   before_action :find_conversation, only: %i[show update answer answer_feedback]
 
   def create
-    conversation = Conversation.new(signon_user: current_user)
+    conversation = Conversation.new(signon_user: current_user, source: :api)
     form = Form::CreateQuestion.new(question_params.merge(conversation:))
 
     if form.valid?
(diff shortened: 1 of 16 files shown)


app/controllers/api/v0/conversations_controller.rb
```diff
@@ -4,7 +4,7 @@ class Api::V0::ConversationsController < ApplicationController
   before_action :find_conversation, only: %i[show update answer answer_feedback]
 
   def create
-    conversation = Conversation.new(signon_user: current_user)
+    conversation = Conversation.new(signon_user: current_user, source: :api)
```
kevindew, line 7: Shall we also add a conditional on `find_conversation` in both web and API so that it only finds them with the appropriate source? I know it seems impossibly edge case that one could migrate to the other but it'd be good to cut off the possibility.

kevindew (commented):
Looks great, just an additional suggestion

app/controllers/api/v0/conversations_controller.rb
davidgisbey: yea good shout. Better safe than sorry 👍

kevindew (approved):
Looks great

PR #212, opened 2025-05-07
Use fulladmin AWS role for bedrock credentials

kevindew:
The govuk-test-poweruser role has been retired and unfortunately the replacement, govuk-test-developer, doesn't have sufficient access that we can invoke models with Bedrock.

Therefore to be able to remain using AWS we need to make use of the fulladmin permission to invoke the model - despite this being advised against.

We're starting a story to update the permissions for the developer role so we can fallback to a reduced access role.


davidgisbey (approved):
lgtm!

PR #214, opened 2025-05-07
Rate limit API endpoints

davidgisbey:
## Description

This adds two throttles:

1. Throttle by token: This will limit the number of requests a single token can make to a specific endpoint within a given time and will protect us being spammed by a single products API consumption
2. Throttle by device id: This will limit the number of requests a single device id can make to a specific endpoint within a given time and will protect us from a single device id spamming the API

I've been generous with the limits for now, but we will probably want to discuss limits with the App team.
I'm going to look into adding rate limit details to the headers in a separate PR/bit of work.

## Tests

I've added to the RackAttack shared examples. Part of me feels like this could just live in one test at the top of the file though rather than testing every endpoint. Or we could move shared functionality (Committee, auth & rate limiting) to an API req spec rather than the conversation controller req spec. Interested to see what people think.

## IP based rate limiting

I've looked into the rate limiting provision Nginx and AWS provide to us as part of being integrated in the GOV.UK Stack. AWS WAF rules allow 2000 requests per ip address every 5 mins. So thats 400 per minute which should be okay for us in the short to mid term. If we got a massive uptick in consumers post rollout we may need to reevaluate due to NATs (Network Address Translation).

Mobile devices often connect through cellular networks that use NATs which can result in many users sharing the same public IP address.

## Additional functionality provided by RackAttack 

Looking into the offering there's stuff they provide which i think we're unlikely to want to use at this point, but may want to down the line

- safelist - app will be public facing so won't need it
- blocklist - may want to use down the line to block malicious actors, but probs overkill for now
  - if we do end up wanting to use this RackAttack has integrated [Fail2Ban](https://github.com/rack/rack-attack?tab=readme-ov-file#fail2ban) and [Allow2Ban](https://github.com/rack/rack-attack?tab=readme-ov-file#fail2ban) which are different strategies for banning malicious users.
- [Tracking users](https://github.com/rack/rack-attack?tab=readme-ov-file#tracks) - we could use this to warn us if API keys are within say 80% of the limit and provide us with info if wanted to, but again seems overkill for now

## Trello card

https://trello.com/c/NGj7izB0/2446-chat-api-rate-limit-api-endpoints

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -41,6 +41,18 @@ class Rack::Attack
     end
   end
 
+  throttle("read requests to Conversations API with device id", limit: 120, period: 1.minute) do |request|
+    if request.path.match?(CONVERSATION_API_PATH_REGEX) && read_method?(request)
+      request.get_header("HTTP_GOVUK_CHAT_CLIENT_DEVICE_ID").presence
+    end
+  end
+
+  throttle("write requests to Conversations API with device id", limit: 20, period: 1.minute) do |request|
+    if request.path.match?(CONVERSATION_API_PATH_REGEX) && !read_method?(request)
+      request.get_header("HTTP_GOVUK_CHAT_CLIENT_DEVICE_ID").presence
+    end
+  end
+
   def self.rails_controller_action(url)
     route = Rails.application.routes.recognize_path(url)
 
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -1,4 +1,6 @@
 class Rack::Attack
+  CONVERSATION_API_PATH_REGEX = /^\/api\/v\d+\/conversation/
+
   throttle("sign-in or sign-ups by IP", limit: 10, period: 5.minutes) do |request|
     homepage_path = Rails.application.routes.url_helpers.homepage_path
     next cdn_client_ip(request) if request.path == homepage_path && request.post?
@@ -27,6 +29,18 @@ class Rack::Attack
     end
   end
 
+  throttle("read requests to Conversations API with token", limit: 10_000, period: 1.minute) do |request|
+    if request.path.match?(CONVERSATION_API_PATH_REGEX) && read_method?(request)
+      normalise_auth_header(request.get_header("HTTP_AUTHORIZATION"))
+    end
+  end
+
+  throttle("write requests to Conversations API with token", limit: 500, period: 1.minute) do |request|
+    if request.path.match?(CONVERSATION_API_PATH_REGEX) && !read_method?(request)
+      normalise_auth_header(request.get_header("HTTP_AUTHORIZATION"))
+    end
+  end
+
   def self.rails_controller_action(url)
     route = Rails.application.routes.recognize_path(url)
 
@@ -40,6 +54,16 @@ class Rack::Attack
     # discriminiator. We can't use request.ip as that uses the IP address of
     # the CDN - so risks blocking the CDN rather than the end user.
     request.get_header("HTTP_TRUE_CLIENT_IP")
+  end
+
+  def self.read_method?(request)
+    request.get? || request.head? || request.options?
+  end
+
+  def self.normalise_auth_header(auth_header)
+    return if auth_header.blank?
+
+    auth_header.strip.gsub(/^bearer/i, "Bearer")
   end
 
   self.throttled_responder = lambda do |request|
--- a/spec/support/rack_attack_examples.rb
+++ b/spec/support/rack_attack_examples.rb
@@ -109,4 +109,55 @@ module RackAttackExamples
       end
     end
   end
+
+  RSpec.shared_examples "throttles traffic for a single device" do |routes:, period:|
+    let(:route_params) { {} }
+    let(:device_id) { "test-device-123" }
+
+    before do
+      read_throttle = Rack::Attack.throttles["read requests to Conversations API with device id"]
+      allow(read_throttle).to receive(:limit).and_return(1)
+
+      write_throttle = Rack::Attack.throttles["write requests to Conversations API with device id"]
+      allow(write_throttle).to receive(:limit).and_return(1)
+    end
+
+    routes.each do |path, methods|
+      methods.each do |method|
+        before do
+          process(method.to_sym,
+                  public_send(path, **route_params),
+                  headers: { "HTTP_GOVUK_CHAT_CLIENT_DEVICE_ID" => device_id })
+        end
+
+        context "when a users device uses its allowance", :rack_attack do
+          it "rejects the next request to #{method} #{path}" do
+            process(method.to_sym,
+                    public_send(path, **route_params),
+                    headers: { "HTTP_GOVUK_CHAT_CLIENT_DEVICE_ID" => device_id })
+
+            expect(response).to have_http_status(:too_many_requests)
+          end
+
+          it "doesn't reject a request with a different device id" do
+            process(method.to_sym,
+                    public_send(path, **route_params),
+                    headers: { "HTTP_GOVUK_CHAT_CLIENT_DEVICE_ID" => "test-device-456" })
+
+            expect(response).not_to have_http_status(:too_many_requests)
+          end
+
+          it "doesn't reject a request after the time period" do
+            travel_to(Time.current + period + 1.second) do
+              process(method.to_sym,
+                      public_send(path, **route_params),
+                      headers: { "HTTP_GOVUK_CHAT_CLIENT_DEVICE_ID" => device_id })
+
+              expect(response).not_to have_http_status(:too_many_requests)
+            end
+          end
+        end
+      end
+    end
+  end
 end
--- a/spec/support/rack_attack_examples.rb
+++ b/spec/support/rack_attack_examples.rb
@@ -1,44 +1,47 @@
 module RackAttackExamples
-  shared_examples "throttles traffic from a single IP address" do |routes:, limit:, period:|
+  RSpec.shared_context "with rack attack helpers" do
+    def process_request(method, path, headers)
+      process(method.to_sym, public_send(path, **route_params), headers: headers)
+    end
+
+    def expect_throttled_response(method, path, headers)
+      process_request(method, path, headers)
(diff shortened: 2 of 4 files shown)


config/initializers/rack_attack.rb
```diff
@@ -27,6 +27,49 @@ class Rack::Attack
...
+    if request.path.match?(/^\/api\/v\d+\/conversation/) && request.get?
+      request.ip
+    end
+  end
+
+  throttle("all other http method requests to Conversations API", limit: 150, period: 1.minute) do |request|
+    if request.path.match?(/^\/api\/v\d+\/conversation/) && !request.get?
+      request.ip
+    end
+  end
```
kevindew, line 40: I notice these are lower than the auth ones - which could be problematic.

I expect we don't actually need an IP one though - I think (but aren't sure - maybe we should check) that there is some basic IP rate limiting that might kick in with our nginx. It feels desirable to try avoid it as it starts to get quite complex to communicate information back.

config/initializers/rack_attack.rb
```diff
@@ -27,6 +27,49 @@ class Rack::Attack
...
+
+  # Other options for throttling
+  ## Throttle by auth token
+  throttle("get requests to Conversations API", limit: 5000, period: 1.minute) do |request|
+    if request.path.match?(/^\/api\/v\d+\/conversation/) && request.get?
+      auth_header = req.env["HTTP_AUTHORIZATION"]
+      token = auth_header.to_s.match(/^Bearer (.+)$/)&.captures&.first
+      token.presence || "no_token_in_header"
```
kevindew, line 48: I imagine we don't want this to fall back to a value as that would push any unauthenticated users into the same rate limit (I notice we've also not got it on the config below)

config/initializers/rack_attack.rb
```diff
@@ -27,6 +27,49 @@ class Rack::Attack
...
+  throttle("all other http method requests to Conversations API", limit: 150, period: 1.minute) do |request|
+    if request.path.match?(/^\/api\/v\d+\/conversation/) && !request.get?
+      request.ip
+    end
+  end
+
+  # Other options for throttling
+  ## Throttle by auth token
+  throttle("get requests to Conversations API", limit: 5000, period: 1.minute) do |request|
+    if request.path.match?(/^\/api\/v\d+\/conversation/) && request.get?
```
kevindew, line 45: shall we put this regex into a constant or perhaps for the match?

kevindew: Oh I expect this should also include HEAD requests as well as GET, maybe OPTIONS too but that's a rarer one - probably something to abstract to a method too though.

config/initializers/rack_attack.rb
```diff
@@ -27,6 +27,49 @@ class Rack::Attack
...
+      auth_header = req.env["HTTP_AUTHORIZATION"]
+      token = auth_header.to_s.match(/^Bearer (.+)$/)&.captures&.first
+      token.presence
+    end
+  end
+
+  ## Throttle by device ID - the app team would have to pass this to us in the header
+  throttle("get requests to Conversations API", limit: 120, period: 1.minute) do |request|
+    if request.path.match?(/^\/api\/v\d+\/conversation/) && request.get?
+      request.get_header("HTTP_X_DEVICE_ID").presence
```
kevindew, line 63: I like this idea 👍 

I think it's now unnecessary to add X to custom headers. I'd wonder if we want it more specific like: HTTP_GOVUK_CHAT_CLIENT_DEVICE_ID ?

kevindew (commented):
This looks great thanks for opening the draft.

I've not done any thought on the actual numbers - but think yours local reasonable and we can chat about them with the app team.

What I think would be useful to look into, and not sure if it's a pain or not is being able to communicate information about the rate limits back in HTTP headers - here's what OpenAI do: https://platform.openai.com/docs/guides/rate-limits#rate-limits-in-headers

I imagine it'd be cool if we could return a header that had a number of entries for account and device limits and where a user is at with them. I have memories that there are means to set that up in rack_attack but where I feel a bit unsure on is if we have the means to say how many are left in a timeframe and how many seconds until the timeframe expires.

config/initializers/rack_attack.rb
davidgisbey: Yea i was mainly looking for a way to stop someone spamming, with no token, but i guess that'd be better covered by an ip address. 

It looks like it won't be necessary anyway if we use a device id which it looks like so far in the consensus 👍

davidgisbey:
Thanks for the feedback. So based on that i'll

- Rate limit based on a device id for now & add a tests for that
- add HEAD and OPTIONS limits
- look into returning useful rate limit info in the headers like OpenAI do

config/initializers/rack_attack.rb
jackbot: Yep this is my preferred option of the three as well. Shouldn't be a problem for the app team to add, and would still be easy to add this header to requests when testing from a local machine.

I guess the only mild concern I have is that if/when we start using the API for the web client then we might need to rethink this as it's obviously not common to include a device ID in the headers of a web client. But I guess that's far enough down the line that we can think about that later.

kevindew: I was expecting that we weren't choosing between options and the proposal was for all 3? Have I misunderstood?

If so I think we'd need both authenticated client and device one, as we want to ensure no signed in service (app) goes crazy but seems good to have something to manage their individual users (devices) too.

davidgisbey: Yea good shout. Down the line we could work around that in various ways if we need to. At the moment it wouldn't rate limit someone if the header wasn't present so we could either tweak this one or add another one for the web client easily enough.

But since we don't really know what's going on with web atm i think we can cross that bridge when we come to it.

davidgisbey: Sure we can do that. In that case i think we should:

- rate limit based on token 
- rate limit based on client id
- rely on Nginx for basic IP rate limiting (i'll look into what that is now)

kevindew: Yup that sounds good 👍

config/initializers/rack_attack.rb
```diff
@@ -27,6 +30,34 @@ class Rack::Attack
     end
   end
 
+  throttle("get requests to Conversations API with token", limit: 10_000, period: 1.minute) do |request|
```
kevindew, line 33:
```suggestion
  throttle("read requests to Conversations API with token", limit: 10_000, period: 1.minute) do |request|
```

tiny nit - just matches your terminology in the code

config/initializers/rack_attack.rb
```diff
@@ -27,6 +30,34 @@ class Rack::Attack
...
+
+  throttle("all other http method requests to Conversations API with token", limit: 500, period: 1.minute) do |request|
+    if request.path.match?(CONVERSATION_API_PATH_REGEX) && !read_method?(request)
+      auth_header = request.env["HTTP_AUTHORIZATION"]
+      token = auth_header.to_s.match(BEARER_TOKEN_REGEX)&.captures&.first
+      token.presence
+    end
+  end
+
+  throttle("get requests to Conversations API with device id", limit: 120, period: 1.minute) do |request|
```
kevindew, line 49:
```suggestion
  throttle("read requests to Conversations API with device id", limit: 120, period: 1.minute) do |request|
```

config/initializers/rack_attack.rb
```diff
@@ -27,6 +30,34 @@ class Rack::Attack
...
+    if request.path.match?(CONVERSATION_API_PATH_REGEX) && read_method?(request)
+      request.get_header("HTTP_GOVUK_CHAT_CLIENT_DEVICE_ID").presence
```
kevindew, line 51: We should document this header in the OpenAPI spec too, not sure if that will be a pain or not as it's used a bunch - hopefully we can use a reference.

I think the way we can do it as flagging an optional header as a parameter https://swagger.io/specification/#parameter-object but that's just from me researching how to do it now so I can't guarantee that won't have problems

config/initializers/rack_attack.rb
```diff
@@ -27,6 +30,34 @@ class Rack::Attack
     end
   end
 
+  throttle("get requests to Conversations API with token", limit: 10_000, period: 1.minute) do |request|
+    if request.path.match?(CONVERSATION_API_PATH_REGEX) && read_method?(request)
+      auth_header = request.env["HTTP_AUTHORIZATION"]
+      token = auth_header.to_s.match(BEARER_TOKEN_REGEX)&.captures&.first
```
kevindew, line 36: I wonder if we're maybe being a bit delicate pulling the actual token out and instead just want to use the whole HTTP_AUTHORIZATION header? I'm not sure if there's any risk that a Bearer token would be accepted if it has say leading or trailing spaces but our regex would reject it for instance. 

If there were any security edge cases to worry about we could always remove whitespace from the value before calling presence

kevindew (commented):
Looks really nice just a few minor things really to look at and maybe discuss.

config/initializers/rack_attack.rb
davidgisbey: ah great catch. Thanks!

config/initializers/rack_attack.rb
davidgisbey: It was pretty straight forward in the end

davidgisbey:
@kevindew thanks for the review. I think it's good for a re-review. No rush at all though!

spec/support/rack_attack_examples.rb
```diff
@@ -1,44 +1,138 @@
...
+          it "normalises Bearer tokens with different formats" do
+            [
+              "bearer testtoken123",
+              "BEARER testtoken123",
+              " Bearer testtoken123",
+              "Bearer testtoken123 ",
+            ].each do |auth_value|
+              process_request(method, path, { "HTTP_AUTHORIZATION" => auth_value })
+              expect(response).to have_http_status(:too_many_requests)
+            end
```
kevindew, line 82: We might well want to keep an eye on whether the tests get a bit slow with things like this running on a number of routes. Could be that it's better to move some of these tests to only test one route and have just a single assertion when applied to each route.

kevindew (approved):
Looks great, just a minor wording suggestion but I don't think this needs another review

PR #220, opened 2025-05-12
Add class for embedding with Titan

jackbot:
Adds a `TextToEmbedding::Titan` class that calls out to Titan via
Bedrock to embed a string or array of strings of text.

Works in exactly the same way that the current OpenAI one does, and the
wrapper `TextToEmbedding` class will call this class if the env var is
set to use Titan as the embedding provider.

Note that this won't really do anything useful yet, until we have the
embeddings in OpenSearch and can query against them. That'll come in the
next PR.

### DIFF
--- a/lib/bedrock_models.rb
+++ b/lib/bedrock_models.rb
@@ -1,3 +1,4 @@
 module BedrockModels
   CLAUDE_3_7_SONNET = "eu.anthropic.claude-3-7-sonnet-20250219-v1:0".freeze
+  TITAN_EMBED_V2 = "amazon.titan-embed-text-v2:0".freeze
 end
--- a/lib/search/text_to_embedding.rb
+++ b/lib/search/text_to_embedding.rb
@@ -4,6 +4,8 @@ module Search
       case llm_provider.to_sym
       when :openai
         Search::TextToEmbedding::OpenAI.call(single_or_collection_of_text)
+      when :titan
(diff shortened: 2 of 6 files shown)


kevindew (commented):
Looks good, do we know what happens if we provide text that is too long to titan? Not sure if we need to consider handling?

kevindew:
> @kevindew yeah good shout. It's a 50k character limit and the API does respond with an exception:
> 
> ```
> Aws::BedrockRuntime::Errors::ValidationException: Malformed input request: expected maxLength: 50000, actual: 160000, please reformat your input and try again. (Aws::BedrockRuntime::Errors::ValidationException)
> ```
> 
> For now I've truncated the input text to 50k characters (like we do with the OpenAI client, except that's with tokens). I figure that's probably sufficient for now. WDYT?

Yeah that sounds good.

I feel ideally we probably want some sort of papertrail that something is truncated - but probably not something right now. Not sure if warning or adding a boolean attribute to search data would be good. Doesn't feel like current approach is right for either of those though.

Shall we maybe add a comment to the code as a note that we may want to improve it. Something like:

```
# For strings longer than the embedding model limit we have to truncate
# the text.
# This is done silently and in future we may want to log this in a
# database or log a warning.
to_embed = text_collection.map(&method(:keep_input_within_text_limit))
```

kevindew (approved):

PR #221, opened 2025-05-12
Accept answer_strategy argument for evaluation:generate_answer

kevindew:
This builds on https://github.com/alphagov/govuk-chat/pull/198 to fix a separate rake task that was also only possible to use the database default of the answer_strategy.

I've configured this task to now accept an answer_strategy argument so a user can specify the answer strategy at runtime.

I thought it was a bit delicate relying on env var for this as it seemed quite easy that a user could be surprised by the answer_strategy used. Therefore I've made this an individual rake argument and set a warning if it's not used to specify what was used.

Following this change we should update the data science documentation to suggest using:

```
bundle exec rake evaluation:generate_answer[claude_structured_answer] QUESTION="what are the vat rates?"
```

over:

```
bundle exec rake evaluation:generate_answer QUESTION="what are the vat rates?" ANSWER_STRATEGY=claude_structured_answer
```

### DIFF
--- a/lib/tasks/evaluation.rake
+++ b/lib/tasks/evaluation.rake
@@ -36,10 +36,13 @@ namespace :evaluation do
   end
 
   desc "Generate a single answer to a question returned as JSON, for 3rd party evaluation tools"
-  task generate_answer: :environment do
+  task :generate_answer, %i[answer_strategy] => :environment do |_, args|
     raise "requires a QUESTION env var" if ENV["QUESTION"].blank?
 
-    question = Question.new(message: ENV["QUESTION"], conversation: Conversation.new)
+    answer_strategy = args.fetch(:answer_strategy, Rails.configuration.answer_strategy)
+    warn "No answer strategy argument provided, using #{answer_strategy}" unless args[:answer_strategy]
+
+    question = Question.new(message: ENV["QUESTION"], conversation: Conversation.new, answer_strategy:)
     answer = AnswerComposition::Composer.call(question)
     puts({ message: answer.message }.to_json)
   end
--- a/spec/lib/tasks/evaluation_spec.rb
+++ b/spec/lib/tasks/evaluation_spec.rb
@@ -96,17 +96,48 @@ RSpec.describe "rake evaluation tasks" do
 
     it "outputs the answer as JSON to stdout" do
       answer = build(:answer)
+      question = "What is the current VAT rate?"
+      answer_strategy = "claude_structured_answer"
 
       allow(AnswerComposition::Composer)
         .to receive(:call)
         .with(an_instance_of(Question))
         .and_return(answer)
 
-      ClimateControl.modify(QUESTION: "What is the current VAT rate?") do
+      ClimateControl.modify(QUESTION: question) do
         answer_json = { message: answer.message }.to_json
-        expect { Rake::Task[task_name].invoke }
+        expect { Rake::Task[task_name].invoke(answer_strategy) }
           .to output("#{answer_json}\n").to_stdout
       end
+
+      expect(AnswerComposition::Composer)
+        .to have_received(:call)
+        .with(an_object_having_attributes(message: question,
+                                          conversation: an_instance_of(Conversation),
+                                          answer_strategy: answer_strategy))
+    end
+
+    it "warns when an answer_strategy argument isn't given" do
+      answer = build(:answer)
+      question = "What is the current VAT rate?"
+      default_answer_strategy = Rails.configuration.answer_strategy
+
+      allow(AnswerComposition::Composer)
+        .to receive(:call)
+        .with(an_instance_of(Question))
+        .and_return(answer)
+
+      ClimateControl.modify(QUESTION: question) do
+        expect { Rake::Task[task_name].invoke }
+          .to output.to_stdout
+          .and output("No answer strategy argument provided, using #{default_answer_strategy}\n").to_stderr
+      end
+
+      expect(AnswerComposition::Composer)
+        .to have_received(:call)
+        .with(an_object_having_attributes(message: question,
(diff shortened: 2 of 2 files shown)


exfalsoquodlibet:
Thanks Kev. I tested this. And I get the following:

- Testing without passing any argument
```
exec rake evaluation:generate_answer QUESTION="i want to start a business but i’m not sure where to get started what’s the cost of getting a business license"
```
returns
```
No answer strategy argument provided, using claude_structured_answer
{"message":"Sorry, something went wrong while trying to answer your question. Try again later.\n\nWe saved your conversation. Check [GOV.UK guidance for businesses](https://www.gov.uk/browse/business) if you need information now.\n"}
```
which suggests the change worked but something else went wrong in the system (this is the standard reply we sent when encountering a system or API error)

- Testing by passing the argument
```
bundle exec rake evaluation:generate_answer[claude_structured_answer] QUESTION="i want to start a business but i’m not sure where to get started what’s the cost of getting a business license"
```
returns:
```
zsh: no matches found: evaluation:generate_answer[claude_structured_answer]
```

kevindew:
> Thanks Kev. I tested this. And I get the following:
> 
> * Testing without passing any argument
> 
> ```
> bundle  exec rake evaluation:generate_answer QUESTION="i want to start a business but i’m not sure where to get started what’s the cost of getting a business license"
> ```
> 
> returns
> 
> ```
> No answer strategy argument provided, using claude_structured_answer
> {"message":"Sorry, something went wrong while trying to answer your question. Try again later.\n\nWe saved your conversation. Check [GOV.UK guidance for businesses](https://www.gov.uk/browse/business) if you need information now.\n"}
> ```
> 
> which suggests the change worked but something else went wrong in the system (this is the standard reply we sent when encountering a system or API error)
> 
> * Testing by passing the argument
> 
> ```
> bundle exec rake evaluation:generate_answer[claude_structured_answer] QUESTION="i want to start a business but i’m not sure where to get started what’s the cost of getting a business license"
> ```
> 
> returns:
> 
> ```
> zsh: no matches found: evaluation:generate_answer[claude_structured_answer]
> ```

Thanks for taking a look.

So I imagine the first issue was caused by a lack of API credentials for AWS - I imagine if you tried it again after running ./bin/export-aws-credentials you'd have success (though of course there are other reasons it could error).

For the second issue, yeah that's a [quirk with zsh that it thinks rake task calls are a zsh function](https://meta.discourse.org/t/solving-zsh-no-matches-found-error-when-running-rake-task/148011) (a regex maybe) by default. I have something in my zsh config that sorts it - I'd actually forgotten about it. If you put the task in quotes you'll not have the issue:

> ```
> bundle exec rake "evaluation:generate_answer[claude_structured_answer]" QUESTION="i want to start a business but i’m not sure where to get started what’s the cost of getting a business license"
> ```

An aside regarding the error and how this task was forgotten:

So this task is one we've not really done anything with for a while as it produces an answer from the full pipeline (guardrails, error handling etc) which I was under the impression we'd not really use anymore with us instead focussing on the task to [generate a structured answer directly](https://github.com/alphagov/govuk-chat/blob/70d752332d7d3aacac28293a461a7b5cb7ac0950/lib/tasks/evaluation.rake#L73) which does just the search and structured answer part and will blow up in a more explanatory way if there is an error. If this task does become more of a key part in the evaluation arsenal we may want to look at making it easier to debug but I realise it also could be that you'd be happy using the other one but we've not really explained how or even that it exists. So may be something to chat over?

lib/tasks/evaluation.rake
```diff
@@ -36,10 +36,17 @@ namespace :evaluation do
...
-  task generate_answer: :environment do
+  task :generate_answer, %i[answer_strategy] => :environment do |_, args|
     raise "requires a QUESTION env var" if ENV["QUESTION"].blank?
 
-    question = Question.new(message: ENV["QUESTION"], conversation: Conversation.new)
+    answer_strategy = args[:answer_strategy]
+    unless answer_strategy
+      answer_strategy_default = Rails.configuration.answer_strategy
+      warn "No answer strategy argument provided, using #{answer_strategy_default}"
+      answer_strategy = answer_strategy_default
```
davidgisbey, line 46: do we need to assign this to a new variable here? We could just pass in 

```
question = Question.new(message: ENV["QUESTION"], conversation: Conversation.new, answer_strategy: answer_strategy_default)
```

Also v nitty feels like `answer_strategy_default` could be `default_answer_strategy`. Especially since that's what the variable is in the test.

kevindew: It's not a new variable as we've already got answer_strategy defined on line 42.

potentially could be more concise though:

```
answer_strategy = args.fetch(:answer_strategy, Rails.configuration.answer_strategy)
unless args[:answer_strategy]
  warn "No answer strategy argument provided, using #{answer_strategy}"
end
```

not sure if that's cryptic at all though, what do you think? 

(btw good shout on answer_strategy_default => default_answer_strategy if we need it)

davidgisbey: ooo i actually really like that 👍

davidgisbey: oh right yea sorry i see you need to pass that variable in when it's assigned so have to overwrite. I like the above but looks good other than using `default_answer_strategy` which is minor so i'll approve so it's unblocked

exfalsoquodlibet:
Thanks @kevindew, this seems to work now for claude (and apologies, my bad for getting the aws credentials step).

I also tested for openai, and `"openai_structured_answer"` also worked.

lib/tasks/evaluation.rake
kevindew: Thanks, just pushed that change and will merge when green

PR #225, opened 2025-05-13
Prototype

JamesCGDS:
Raising a draft PR to get this deployed to the preview app.


.gitignore
```diff
@@ -72,3 +72,9 @@ yarn-debug.log*
 # Ignore assets compiled for dart-sass
 /app/assets/builds/*
 !/app/assets/builds/.keep
+
+.DS_Store
+app/.DS_Store
+app/assets/.DS_Store
+app/assets/images/.DS_Store
+app/assets/images/components/.DS_Store
```
kevindew, line 80: FYI these are best ignored on your own system with a gitignore global file rather than putting it into the application: https://docs.github.com/en/get-started/git-basics/ignoring-files#excluding-local-files-without-creating-a-gitignore-file

Also you should only need .DS_Store in once and it'd ignore it in all directories.

JamesCGDS: Cool, didn't know that re `.git/info/exclude` - thanks

PR #226, opened 2025-05-13
Titan embedding and retrieval

jackbot:
* Adds a mapping in OpenSearch for `titan_embedding` to allow us to store the embedding data against a document. We can then use this field when querying the index
* Updates the seed task to generate embeddings for Titan as well as OpenAI
* When pulling content items from the GOV.UK message queue, we generate and store the Titan embeddings

Note that we'll continue to keep the OpenAI embeddings around and generate embeddings for both providers for all documents in the store. This allows us to switch between providers for testing and evaluation.

At some point we'll be removing the OpenAI embeddings once we decide on a replacement provider. For that reason there's quite a bit of code duplication which would be nicer to tidy up (e.g. in a loop to generate both provider embeddings), but given we'll be removing OpenAI, the duplication won't be around for long.

### DIFF
--- a/lib/search/chunked_content_repository.rb
+++ b/lib/search/chunked_content_repository.rb
@@ -1,6 +1,7 @@
 module Search
   class ChunkedContentRepository
     OPENAI_EMBEDDING_DIMENSIONS = 3072
+    TITAN_EMBEDDING_DIMENSIONS = 1024
     MAPPINGS = {
       content_id: { type: "keyword" },
       locale: { type: "keyword" },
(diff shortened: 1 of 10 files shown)


lib/search/chunked_content_repository.rb
```diff
@@ -23,6 +24,15 @@ class ChunkedContentRepository
...
       },
+      titan_embedding: {
+        type: "knn_vector",
+        dimension: TITAN_EMBEDDING_DIMENSIONS,
+        method: {
+          name: "hnsw",
+          space_type: "cosinesimil",
+          engine: "nmslib",
+        },
+      },
```
kevindew, line 35: I have a recollection that we have a method to update mappings, but not sure we ever used it in anger. Have you looked into this? I guess we have some sort of task to run - or worse, code to write

kevindew (commented):
This looks great - I do think before merging we should have a chat with @ianhowell-gds to get his opinion on what actions we should take (if any) for how this will affect our OpenSearch instances

PR #231, opened 2025-05-14
Return rate limits in headers

davidgisbey:
## Description 

We want to implement adding useful information on our rate limits in the headers of the response we send back to API consumers. They can use this to debug, track stats on dashboards etc.

This PR builds on #214 which adds throttles to our API endpoints by providing this functionality. 

While RackAttack doesn't provide this functionality our the box they populate the env with the following:

```
env["rack.attack.throttle_data"]

{
  "read requests to Conversations API with device id" =>
  {
    discriminator: "test-device-123",
    count: 2,
    period: 60,
    limit: 1,
    epoch_time: 1747308452
  }
}
```

This means we can just grab the info on each throttle and use it to populate the rate limit headers.

As we will need to be able to send info back about multiple throttles (token and device id) i've prepended the header with `Token` or `DeviceId` to make it clear which limit applies to which throttle. I've also prepended write or read so it's easy to distinguish between the throttles that cover each.

I've updated the OpenAPI spec to cover the new functionality.

## Headers in postman 

image

## Trello card

https://trello.com/c/NGj7izB0/2446-chat-api-rate-limit-api-endpoints


config/initializers/rack_attack.rb
```diff
@@ -1,5 +1,46 @@
 class Rack::Attack
   CONVERSATION_API_PATH_REGEX = /^\/api\/v\d+\/conversation/
+  TOKEN_READ_THROTTLE_NAME = "read requests to Conversations API with token".freeze
+  TOKEN_WRITE_THROTTLE_NAME = "write method requests to Conversations API with token".freeze
+  DEVICE_READ_THROTTLE_NAME = "read requests to Conversations API with device id".freeze
+  DEVICE_WRITE_THROTTLE_NAME = "write requests to Conversations API with device id".freeze
+
+  class RateLimitHeadersMiddleware
```
kevindew, line 8: Is it worth moving this to lib so we can unit test it directly? with the usual caveat that testing rack middleware is a bit of a pain.

config/initializers/rack_attack.rb
```diff
@@ -1,5 +1,46 @@
...
+    THROTTLE_TO_PREFIX_MAPPING = {
+      TOKEN_READ_THROTTLE_NAME => "Token-Read",
+      TOKEN_WRITE_THROTTLE_NAME => "Token-Write",
```
kevindew, line 11: I wonder about going for API User rather than token here? it feels a bit more intuitive

config/initializers/rack_attack.rb
```diff
@@ -1,5 +1,46 @@
...
+      DEVICE_READ_THROTTLE_NAME => "Device-Id-Read",
+      DEVICE_WRITE_THROTTLE_NAME => "Device-Id-Write",
```
kevindew, line 13: I think we may want to try have a bit of consistency in how we name these and how we expect the client device header to be `Govuk-Chat-Client-Device-Id`.

So if we want to keep the Govuk-Chat prefix perhaps these become:

`Govuk-Chat-Api-User-Read-*`
`Govuk-Chat-Api-User-Write-*`
`Govuk-Chat-Client-Device-Id-Read-*`
`Govuk-Chat-Client-Device-Id-Write-*`

These do end up being a bit long so we may want to find a way of shortening them, so we could drop chat from the name and id so they are:

`Govuk-Api-User-Read-*`
`Govuk-Api-User-Write-*`
`Govuk-Client-Device-Read-*`
`Govuk-Client-Device-Write-*`

kevindew (commented):
Sorry ran out of time, so only a super partial review. Looks really cool though, nice one for working this all out.

kevindew:
Will look again tomorrow

config/initializers/rack_attack.rb
davidgisbey: yea i initially added this to lib. I'll move it back and add some tests 👍

config/initializers/rack_attack.rb
davidgisbey: also makes sense

lib/api/rate_limit_middleware.rb
```diff
@@ -0,0 +1,39 @@
+class Api::RateLimitMiddleware
+  GOVUK_CHAT_API_USER_READ_THROTTLE_NAME = "read requests to Conversations API with token".freeze
```
davidgisbey, line 2: it feels a bit off to pull these constants into the RackAttack initialiser for the throttle names. They more Maybe they could live in `lib/api/rate_limit_constants.rb` or similar? Wdyt?

kevindew: Yeah sounds good to move them out. I'm not so keen on a name that explicitly mentions constants. Maybe we want to have 

Api::RateLimit (constants)
then Api::RateLimit::Middleware?

davidgisbey:
So i've managed to track down that we can get the exact seconds until the throttle runs out https://github.com/rack/rack-attack/blob/main/lib/rack/attack/cache.rb#L69C1-L75C1 & https://github.com/rack/rack-attack/blob/main/lib/rack/attack/cache.rb#L69C1-L75C1 

by doing 

```
epoch_time = throttle_info[:epoch_time]
expires_in = (period - (epoch_time % period) + 1).to_i
```

I'm getting failures in my tests. It looks like there's some state leaking between tests. It's returning the seconds between when the cache was first hit in a test (starts at 60) and counts down from there. Even when i call `Rack::Attack.reset!` as suggested in the docs.

We're creating a [new cache memory store for each test though](https://github.com/alphagov/govuk-chat/blob/9e62629a7c8a850fe0a1d8017877b212ccd7a3ff/spec/spec_helper.rb#L76-L81) too which seemed like the other fix 🤔

lib/api/rate_limit_middleware.rb
davidgisbey: I've done this. As there's a bunch of prior art for it i've just nested them in the same file rather than explicitly pulling them into a separate file. Although happy to that if you think it's prerable

lib/api/rate_limit.rb
```diff
@@ -0,0 +1,42 @@
...
+      throttle_data = env["rack.attack.throttle_data"]
+
+      if throttle_data
+        throttle_data.each do |throttle_name, throttle_info|
+          header_name_prefix = THROTTLE_TO_PREFIX_MAPPING[throttle_name]
+          limit = throttle_info[:limit]
+          count = throttle_info[:count]
+          period = throttle_info[:period]
+          epoch_time = throttle_info[:epoch_time]
+          expires_in = (period - (epoch_time % period)).to_i
```
davidgisbey, line 31: i suspect you already know all about epoch time, but i wasn't familiar.

Essentially what it does in case anyone is looking and interested, is it counts all the seconds that have passed since 1/1/1970. Then divides it by the period (60 in our case). Any remainder is how many seconds have passed since the beginning of the current period.

So say the remainder is 10, it'll be 50 secs until reset as we can just do 60 (our period) - 10 (the remainder)

davidgisbey:
@kevindew thanks for the review. I think it's ready for a re-review 👍

lib/api/rate_limit.rb
kevindew: Great thanks for explaining. I didn't know it worked like that either. I can see some of the logic of it - having these fixed time windows makes it all much easier to reason on and can probably make things much simpler if say having x requests an hour actually changes when the hour changes rather than an arbitrary time within it. 

But yes definitely a bit painful that someone could spam you at the end of a period with the whole time limits amount (which admittedly could still be a case for a rolling window if someone was allowed 500 requests in 5 mins and did 1 request, waited 4:58 seconds and then hit you with the last 499)

kevindew (commented):
Looks great David - have unfortunately given you a bunch of annoying OpenAPI stuff - maybe we'll have to hook this up to the gem I wrote to validate it as I got loads trying it out:

```
irb(main):003:0> o3p = Openapi3Parser.load_file("/Users/kevin.dew/govuk/govuk-chat/docs/api_openapi_speci
fication.yml")
=> Openapi3Parser::Document(openapi_version: 3.0, root_source: Openapi3Parser::Source(input: /Users/...
irb(main):004:0> o3p.errors
=> Openapi3Parser::Validation::ErrorCollection(errors: {"#/paths/~1conversation/post/responses/201/headers"=>["Invalid type. Expected Object"], "#/paths/~1conversation/post/responses/422/headers"=>["Invalid type. Expected Object"], "#/paths/~1conversation/post/responses/429/%24ref"=>["#/components/responses/TooManyWriteRequests does not resolve to a valid object"], "#/paths/~1conversation~1%7Bconversation_id%7D/get/responses/200/headers"=>["Invalid type. Expected Object"], "#/paths/~1conversation~1%7Bconversation_id%7D/get/responses/404/headers"=>["Invalid type. Expected Object"], "#/paths/~1conversation~1%7Bconversation_id%7D/get/responses/429/%24ref"=>["#/components/responses/TooManyReadRequests does not resolve to a valid object"], "#/paths/~1conversation~1%7Bconversation_id%7D/put/responses/201/headers"=>["Invalid type. Expected Object"], "#/paths/~1conversation~1%7Bconversation_id%7D/put/responses/422/headers"=>["Invalid type. Expected Object"], "#/paths/~1conversation~1%7Bconversation_id%7D/put/responses/429/%24ref"=>["#/components/responses/TooManyWriteRequests does not resolve to a valid object"], "#/paths/~1conversation~1%7Bconversation_id%7D~1questions~1%7Bquestion_id%7D~1answer/get/responses/200/headers"=>["Invalid type. Expected Object"], "#/paths/~1conversation~1%7Bconversation_id%7D~1questions~1%7Bquestion_id%7D~1answer/get/responses/202/headers"=>["Invalid type. Expected Object"], "#/paths/~1conversation~1%7Bconversation_id%7D~1questions~1%7Bquestion_id%7D~1answer/get/responses/404/headers"=>["Invalid type. Expected Object"], "#/paths/~1conversation~1%7Bconversation_id%7D~1questions~1%7Bquestion_id%7D~1answer/get/responses/429/%24ref"=>["#/components/responses/TooManyReadRequests does not resolve to a valid object"], "#/paths/~1conversation~1%7Bconversation_id%7D~1answers~1%7Banswer_id%7D~1feedback/post/responses/201/headers"=>["Invalid type. Expected Object"], "#/paths/~1conversation~1%7Bconversation_id%7D~1answers~1%7Banswer_id%7D~1feedback/post/responses/422/headers"=>["Invalid type. Expected Object"], "#/paths/~1conversation~1%7Bconversation_id%7D~1answers~1%7Banswer_id%7D~1feedback/post/responses/404/headers"=>["Invalid type. Expected Object"], "#/paths/~1conversation~1%7Bconversation_id%7D~1answers~1%7Banswer_id%7D~1feedback/post/responses/429/%24ref"=>["#/components/responses/TooManyWriteRequests does not resolve to a valid object"], "#/components/responses/TooManyReadRequests/headers"=>["Invalid type. Expected Object"], "#/components/responses/TooManyWriteRequests/headers"=>["Invalid type. Expected Object"], "#/components/headers/ReadRateLimitHeaders"=>["Unexpected fields: Govuk-Chat-Api-User-Read-RateLimit-Limit, Govuk-Chat-Api-User-Read-RateLimit-Remaining, Govuk-Chat-Api-User-Read-RateLimit-Reset, Govuk-Chat-Client-Device-Id-Read-RateLimit-Limit, Govuk-Chat-Client-Device-Id-Read-RateLimit-Remaining and Govuk-Chat-Client-Device-Id-Read-RateLimit-Reset"], "#/components/headers/WriteRateLimitHeaders"=>["Unexpected fields: Govuk-Chat-Api-User-Write-RateLimit-Limit, Govuk-Chat-Api-User-Write-RateLimit-Remaining, Govuk-Chat-Api-User-Write-RateLimit-Reset, Govuk-Chat-Client-Device-Id-Write-RateLimit-Limit, Govuk-Chat-Client-Device-Id-Write-RateLimit-Remaining and Govuk-Chat-Client-Device-Id-Write-RateLimit-Reset"]})
```

I did try run it locally against Heroku and it seems that without having a bearer token I'm still getting the rate limit for that returned (and decreasing) so I think there's perhaps an issue with how that is recorded (I guess that might be a separate PR)

```
➜  ~ curl -I -u $credentials -H 'Govuk_Chat_Client_Device_Id: 123' https://govuk-chat-2446-return--x0ht0p.herokuapp.com/api/v0/conversation/df188bc5-273a-45c9-b386-657a763b61d1 -s | grep 'Govuk'
Govuk-Chat-Api-User-Read-Ratelimit-Limit: 10000
Govuk-Chat-Api-User-Read-Ratelimit-Remaining: 9996
Govuk-Chat-Api-User-Read-Ratelimit-Reset: 9s
Govuk-Chat-Client-Device-Id-Read-Ratelimit-Limit: 120
Govuk-Chat-Client-Device-Id-Read-Ratelimit-Remaining: 116
Govuk-Chat-Client-Device-Id-Read-Ratelimit-Reset: 9s
➜  ~ curl -I -u $credentials https://govuk-chat-2446-return--x0ht0p.herokuapp.com/api/v0/conversation/df188bc5-273a-45c9-b386-657a763b61d1 -s | grep 'Govuk'
Govuk-Chat-Api-User-Read-Ratelimit-Limit: 10000
Govuk-Chat-Api-User-Read-Ratelimit-Remaining: 9999
Govuk-Chat-Api-User-Read-Ratelimit-Reset: 55s
```

Edit: I've just realised why we get the api user header it's because I'm using basic auth for Heroku and there will indeed be an Authorization header set with basic: `Authorization: Basic <base64>`

config/initializers/rack_attack.rb
davidgisbey: oh right yea sorry sorting that now

spec/lib/api/rate_limit_spec.rb
```diff
@@ -0,0 +1,61 @@
+RSpec.describe Api::RateLimit do
+  describe Api::RateLimit::Middleware do
+    let(:app_double) do
+      instance_double(
+        ActiveSupport::Cache::Strategy::LocalCache::Middleware,
+        call: [200, { "Content-Type" => "application/json" }, []],
+      )
```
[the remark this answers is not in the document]
davidgisbey: ypu def can pass a lambda

docs/api_openapi_specification.yml
```diff
@@ -33,13 +33,17 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/PendingQuestion"
+          headers:
+            - $ref: "#/components/headers/WriteRateLimitHeaders"
```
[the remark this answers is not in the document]
davidgisbey: ah damn that's how i started if off, but it makes the file incredibly repetitive. Oh well can't be avoided i'll sort that now

lib/api/rate_limit.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    def call(env)
+      status, headers, body = @app.call(env)
+
+      throttle_data = env["rack.attack.throttle_data"]
+
+      if throttle_data
+        throttle_data.each do |throttle_name, throttle_info|
+          header_name_prefix = THROTTLE_TO_PREFIX_MAPPING[throttle_name]
```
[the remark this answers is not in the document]
davidgisbey: Okay i'll add something for this. I think in this situation it's probs best to just log it so we don't stop valid requests being blocked. Wdyt?

docs/api_openapi_specification.yml
```diff
@@ -212,6 +236,93 @@ components:
         consume all of an API users' limits.
       schema:
         type: string
+  headers:
+    ReadRateLimitHeaders:
+      Govuk-Chat-Api-User-Read-RateLimit-Limit:
+        $ref: '#/components/headers/GovuChatApiUserReadRateLimitLimit'
```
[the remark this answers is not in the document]
davidgisbey: Yea i would've expected committee to blow up on a bunch of this stuff. I'll have a look into why that's not happened.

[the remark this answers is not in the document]
davidgisbey: It looks like Committee only validates that headers that are defined  [adhere to the schema and are present if required](https://github.com/ota42y/openapi_parser/blob/master/lib/openapi_parser/parameter_validator.rb#L8-L25). So it ignores any additional headers that are present in the response. Since the one being returned was different due to the typo it wouldn't have been validated against the schema.

docs/api_openapi_specification.yml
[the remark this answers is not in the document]
kevindew: Yeah sorry, does seem a bit painful. It's not something that is changed on newer version of OpenAPI eithers

lib/api/rate_limit.rb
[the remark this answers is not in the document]
kevindew: Yup sounds a good idea

docs/api_openapi_specification.yml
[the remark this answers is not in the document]
kevindew: ok cool, so with the typo fixed and an issue would it blow up ok?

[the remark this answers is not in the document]
davidgisbey: ergh damn. My bad, that's the parameter validator that only gets used for request headers not response headers. Annoyingly there's no option to configure it for responses like there is for [requests](https://github.com/interagent/committee?tab=readme-ov-file#configuration-options)

So it looks like we either need to:

- find something else to validate response headers 
- build something ourselves to do it
- accept that they aren't being validated

Maybe we can chat about it later if you've got some time. Whatever we do i think we should pick it up as a separate to this PR. Let me know if you're free and/or disagree with that.

davidgisbey:
@kevindew i've addressed the comments, but it looks like there's an issue with validating response headers via committee that i've discussed [here](https://github.com/alphagov/govuk-chat/pull/231#discussion_r2097990565).

kevindew (approved):
Looks great ⭐

PR #237, opened 2025-05-16
Ensure LinkTokenMapper#mapping uses absolute URLs

davidgisbey:
## Description

At the moment we're using relative URLs to store the links to our chunks in the mappings. This means when we convert our links from tokens to the URLs and render them on the page they're relative links and blow up on Heroku since they vist heroku.com/relative-path rather than gov.uk/relative-path.

We do convert relative links to absolute URLs in the #map_links_to_tokens method. This is used to format links within the content of a chunk [during answer composition](https://github.com/alphagov/govuk-chat/blob/960ce6dbb3fabc9aac114933b26b18febfa6ab78/lib/answer_composition/pipeline/context.rb#L79-L81)

But we call [LinkTokenMapper#map_link_to_token directly on the chunks url](https://github.com/alphagov/govuk-chat/blob/960ce6dbb3fabc9aac114933b26b18febfa6ab78/lib/answer_composition/pipeline/context.rb#L75). 

This means that the chunks path was never getting converted to an absolute URL.

By moving the logic to convert relative URLs to absolute URLs into the LinkTokenMapper#map_link_to_token method we can ensure that all links are converted to absolute URLs.


lib/answer_composition/link_token_mapper.rb
```diff
@@ -12,24 +12,23 @@ def map_links_to_tokens(html_content, exact_path)
...
       doc.to_html
     end
 
-    def map_link_to_token(link)
+    def map_link_to_token(link, exact_path = nil)
+      link = begin
+        if exact_path.present?
+          URI.join(ensure_absolute_govuk_url(exact_path), link).to_s
+        else
+          ensure_absolute_govuk_url(link)
```
kevindew, line 26: I think we probably want to rename this method to ensure_absolute_url as it doesn't ensure the URL is a GOV.UK one

lib/answer_composition/link_token_mapper.rb
```diff
@@ -12,24 +12,23 @@ def map_links_to_tokens(html_content, exact_path)
...
-          link["href"] = token
-        end
+        link["href"] = map_link_to_token(link["href"], exact_path)
       end
 
       doc.to_html
     end
 
-    def map_link_to_token(link)
+    def map_link_to_token(link, exact_path = nil)
```
kevindew, line 21: We might want to rename exact_path to base_path here to reflect that it's what we're using as the base of the URL

lib/answer_composition/link_token_mapper.rb
davidgisbey: yup great shout

lib/answer_composition/link_token_mapper.rb
davidgisbey: yea def clearer thanks!

davidgisbey:
@kevindew thanks i've made those changes.

PR #241, opened 2025-05-20
Remove Early Access & Waiting List users from admin homepage

Sunny-xyz:
Trello: https://trello.com/c/zhI0UIU3/2464-remove-earlyaccessuser-and-waitinglistuser-functionality-from-the-admin-area

This PR removes the @early_access_user_stats and @waiting_list_user_stats from the controller, strips out the panels from the view, and updates the homepage request spec, as part of the ongoing effort to remove EarlyAccess & WaitingList uses from the codebase

### DIFF
--- a/app/views/admin/homepage/index.html.erb
+++ b/app/views/admin/homepage/index.html.erb
@@ -29,25 +29,6 @@ end
           This can be changed in <%= link_to("settings", admin_settings_path, class: "govuk-link") %>.
         </p>
       <% end %>
-    <% elsif !settings.sign_up_enabled %>
-      <%= render "govuk_publishing_components/components/notice", {
-        title: "Sign ups are disabled",
-      } do %>
-        <p class="govuk-body">
-          All users who attempt to sign up will be denied. Only existing users
-          can login.
-          <%= link_to("Update settings", admin_settings_path, class: "govuk-link") %>.
-        </p>
-      <% end %>
-    <% elsif @waiting_list_user_stats[:percentage_of_waiting_list_used] >= 80 %>
-      <%= render "govuk_publishing_components/components/notice", {
-        title: "The waiting list is #{number_to_percentage(@waiting_list_user_stats[:percentage_of_waiting_list_used], precision: 0)} full",
-      } do %>
-        <p class="govuk-body">
-          There are <%= @waiting_list_user_stats[:current] %> users on the waiting list. The maximum allowed is <%= settings.max_waiting_list_places %>.
-          If you would like to increase the maximum places you can <%= link_to("update the setting", admin_settings_edit_max_waiting_list_places_path, class: "govuk-link") %>.
-        </p>
-      <% end %>
     <% end %>
   </div>
 </div>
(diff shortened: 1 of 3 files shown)


kevindew:
I can see this is still a WIP but just wanted to provide some feedback on how we can look at breaking this up to keep it all manageable for reviews and spotting any issues

I'd suggest you probably look at having around 1 PR per controller of the admin area, which should all prove to be mostly independent.

So something like:

- Remove early access/waiting list users from admin homepage
- Remove references to early access users from questions / conversations
- Remove early access/waiting list users from metrics
- Remove access controls for Early Access Users
- Remove waiting list user management
- Remove early access user management (likely depends on Remove access controls for Early Access Users)
- Remove early access/waiting list user functionality from settings

I think that would end up with quite manageable PRs but not block you from moving onto another one while one is in review.

app/views/admin/homepage/index.html.erb
```diff
@@ -16,42 +16,6 @@ if ENV["GOVUK_ENVIRONMENT"]
 end
 %>
 
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-full">
-    <% if !settings.public_access_enabled %>
```
kevindew, line 21: This setting will still have an effect when the early access users / waiting list users are removed so we'll need to keep this conditional.

We will at a point in this need to change the copy though so if you could add that to the trello card as a checklist so we don't forget.

app/views/admin/homepage/index.html.erb
```diff
@@ -16,42 +16,6 @@ if ENV["GOVUK_ENVIRONMENT"]
...
-        title: "Public access to chat is disabled",
-      } do %>
-        <%= render "govuk_publishing_components/components/warning_text", {
-          text: "All users attempting to use chat will receive an error response",
-        } %>
-        <p class="govuk-body">
-          This can be changed in <%= link_to("settings", admin_settings_path, class: "govuk-link") %>.
-        </p>
-      <% end %>
-    <% elsif !settings.sign_up_enabled %>
```
kevindew, line 32: This one looks good to remove

app/views/admin/homepage/index.html.erb
```diff
@@ -16,42 +16,6 @@ if ENV["GOVUK_ENVIRONMENT"]
...
-      <%= render "govuk_publishing_components/components/notice", {
-        title: "Sign ups are disabled",
-      } do %>
-        <p class="govuk-body">
-          All users who attempt to sign up will be denied. Only existing users
-          can login.
-          <%= link_to("Update settings", admin_settings_path, class: "govuk-link") %>.
-        </p>
-      <% end %>
-    <% elsif @waiting_list_user_stats[:percentage_of_waiting_list_used] >= 80 %>
```
kevindew, line 42: This looks good to remove.

kevindew (commented):
Looks like this deletes just a bit too much so need to put some code back.

kevindew (approved):
Nice one, this looks good to merge now. Let's just keep a record that we need to change the public_access_enabled copy as we reach the end of this work.

PR #249, opened 2025-05-27
Update pending question blueprint to use external chat url

davidgisbey:
## Description 

Since chat is not longer being hosted on www we need to update the API to return the external chat URL when a question is pending i.e.

`https://chat.gov.uk/api/v0/conversation/{conversation_id}/questions/{question_id}/answer`

rather than

`https://www.gov.uk/api/v0/conversation/{conversation_id}/questions/{question_id}/answer`

I've had a look for other examples that are using `Plek.website_root` where we shouldn't be.

There are a couple:

1. in the [mailer options](https://github.com/alphagov/govuk-chat/blob/ed2aecb8aa25b60020b67bd2a8700836ab7abc8c/config/environments/production.rb#L96)
2. when users unsubscribe we send them an email to [sign up again ](https://github.com/alphagov/govuk-chat/blob/ed2aecb8aa25b60020b67bd2a8700836ab7abc8c/app/views/unsubscribe/waiting_list_user.html.erb#L10)if they change their mind

Both of these are being remove by Chae as part of this [PR](https://github.com/alphagov/govuk-chat/pull/242).

I won't worry about touching them here as it'll create merge conflicts.


kevindew (approved):
Looks good, thanks for this

PR #252, opened 2025-05-28
Remove Early Access/Waiting List Users from Metrics

Sunny-xyz:
WIP


Sunny-xyz:
Changes from this PR that affect the admin metrics page can be observed below:

Before:
image


After:
image

PR #259, opened 2025-05-30
Remove Access Controls for Early Access Users

Sunny-xyz:
WIP


app/helpers/admin/early_access_users_helper.rb
```diff
@@ -1,23 +1,7 @@
...
-                  else
-                    ""
-                  end
-
-    safe_join([
-      link_to(user_email, admin_early_access_user_path(user), class: "govuk-link"),
-      user_suffix,
-    ])
+    user_email = user.email
+    link_to(user_email, admin_early_access_user_path(user), class: "govuk-link")
```
davidgisbey, line 4: probs can just update this to 

```
link_to(user_email, admin_early_access_user_path(user), class: "govuk-link")
```

in fact i think we can just strip this out the helper now since it's really not worth unit testing creating a basic link.

So we can just pop this link [here](https://github.com/alphagov/govuk-chat/blob/58fe038dcaf390869c768d40c6c67d23048fba37/app/views/admin/early_access_users/_users_table.html.erb#L30) and remove the helper method and tests

app/models/admin/filters/early_access_users_filter.rb
```diff
@@ -45,10 +45,6 @@ def access_scope(scope)
     return scope if access.blank?
 
     case access
-    when "revoked"
-      scope.where.not(revoked_at: nil)
-    when "shadow_banned"
-      scope.where.not(shadow_banned_at: nil)
     when "at_question_limit"
```
davidgisbey, line 48: generally you wouldn't use a case statement when there's only 2 options. You'd use a conditional instead

```
 if access == "at_question_limit"
  scope.at_question_limit
else
  scope.where(revoked_at: nil, shadow_banned_at: nil).within_question_limit
end
```

But since you're stripping it out in the next PR dw about it for this one

davidgisbey (commented):
Just one minor comment. Other than that it looks good

app/helpers/admin/early_access_users_helper.rb
Sunny-xyz: Thanks for this - I've extracted the links from the helper & placed them in the user table file now, and removed the original helper + its spec file.

PR #261, opened 2025-06-02
Trigger guardrails when JailbreakChecker returns unexpected result 

davidgisbey:
## Description 

With Claude we don’t have the ability we had with OpenAI to specify that a model can only return a single token from a limited list of options (thus our ability to force a 0 or 1 response).

We therefore want to avoid a situation where the system raises an exception when the model reaches the max tokens limit, and passes back an unexpected value in the response for `llm_guardrail_result`.

We want to be able to see the response in full to see what the issue is so we've bumped the max_tokens to 50 in the prompt config in this [PR](https://github.com/alphagov/govuk_chat_private/pull/20).

This updates the jailbreak checker to only raise an exception for non-claude models.

## Trello card

https://trello.com/c/oMWS5Kgp/2511-change-jailbreak-guardrails-to-be-more-forgiving-for-claude

### DIFF
--- a/lib/guardrails/jailbreak_checker.rb
+++ b/lib/guardrails/jailbreak_checker.rb
@@ -30,10 +30,6 @@ module Guardrails
       guardrails_llm_prompts.fetch(:pass_value)
     end
 
-    def self.fail_value
-      guardrails_llm_prompts.fetch(:fail_value)
-    end
-
     def self.guardrails_llm_prompts
       Rails.configuration.govuk_chat_private.llm_prompts.common.jailbreak_guardrails
     end
@@ -55,17 +51,10 @@ module Guardrails
         raise "Unexpected provider #{llm_provider}"
       end
 
-      case result[:llm_guardrail_result]
-      when fail_value
-        create_result(result, triggered: true)
-      when pass_value
+      if result[:llm_guardrail_result] == pass_value
         create_result(result, triggered: false)
       else
-        raise ResponseError.new(
-          "Error parsing jailbreak guardrails response",
-          llm_guardrail_result: result[:llm_guardrail_result],
-          **result_attributes(result),
-        )
+        create_result(result, triggered: true)
       end
     end
 
@@ -73,7 +62,7 @@ module Guardrails
 
     attr_reader :input, :llm_provider
 
-    delegate :guardrails_llm_prompts, :pass_value, :fail_value, to: :class
+    delegate :guardrails_llm_prompts, :pass_value, to: :class
 
     def create_result(result, triggered:)
       Result.new(
--- a/spec/lib/guardrails/jailbreak_checker_spec.rb
+++ b/spec/lib/guardrails/jailbreak_checker_spec.rb
@@ -1,10 +1,9 @@
 RSpec.describe Guardrails::JailbreakChecker do
   let(:input) { "User question" }
   let(:pass_value) { "PassValue" }
-  let(:fail_value) { "FailValue" }
 
   before do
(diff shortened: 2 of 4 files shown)


lib/guardrails/jailbreak_checker.rb
```diff
@@ -55,11 +55,12 @@ def call
...
 
-      case result[:llm_guardrail_result]
-      when fail_value
-        create_result(result, triggered: true)
-      when pass_value
+      guardrail_result = result[:llm_guardrail_result]
+
+      if guardrail_result == pass_value
         create_result(result, triggered: false)
+      elsif guardrail_result == fail_value || llm_provider == :claude
```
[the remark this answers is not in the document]
davidgisbey: okay sure sounds good to me

davidgisbey:
@kevindew thanks i've made that change

lib/guardrails/jailbreak_checker.rb
```diff
@@ -56,16 +56,10 @@ def call
...
-        create_result(result, triggered: true)
       when pass_value
         create_result(result, triggered: false)
       else
-        raise ResponseError.new(
-          "Error parsing jailbreak guardrails response",
-          llm_guardrail_result: result[:llm_guardrail_result],
-          **result_attributes(result),
-        )
+        create_result(result, triggered: true)
```
kevindew, line 62: I think we'll just have some cleaning up left about removing fail_value as a method from this, the corresponding class, the tests and the prompt

spec/lib/guardrails/jailbreak_checker_spec.rb
```diff
@@ -23,8 +23,6 @@
     }
 
     allow(Guardrails::OpenAI::JailbreakChecker).to receive(:call).and_return(result)
-    stub_openai_jailbreak_guardrails(input)
```
kevindew, line 26: Nice catch

lib/guardrails/jailbreak_checker.rb
```diff
@@ -56,16 +56,10 @@ def call
       end
 
       case result[:llm_guardrail_result]
-      when fail_value
-        create_result(result, triggered: true)
       when pass_value
```
kevindew, line 59: it seems a bit suprising to use a case now it's just a single conditional

spec/lib/guardrails/jailbreak_checker_spec.rb
```diff
@@ -76,55 +76,21 @@
...
-    end
-
-    it "raises an error that can be represented as JSON" do
-      error = nil
-      begin
-        described_class.call(input)
-      rescue described_class::ResponseError => e
-        error = e
-      end
+  it "returns a result object with triggered true when llm_guardrail_result is an unexpected value" do
```
kevindew, line 75: I'm not sure we'll really have this as a third option test anymore and expect it'll end up merged with the test above as the code is no longer pass_value, fail_value or unexpected - as fail and unexpected should be one and the same

kevindew (commented):
Thanks for this, I think there's a few more bits we can do to get this a bit more cleanly integrated.

spec/lib/guardrails/jailbreak_checker_spec.rb
davidgisbey: Cool i've merged the tests to document that we treat both fails and unexpected values in the same way. I can see the argument for just testing one (probably fail since it's more common. Let me know if you'd like me to strip it out.

davidgisbey:
@kevindew i've made those changes. I've also opened a PR on the private repo to remove the fail_value https://github.com/alphagov/govuk_chat_private/pull/21

PR #274, opened 2025-06-04
Limit and only return active questions in API

davidgisbey:
## Description

The OpenAPI specification states

```
Returns a list of AnsweredQuestions with potentially one PendingQuestion.
Is limited to returning 500 questions for a conversation and only
questions asked within last 30 days.
```

I missed this during the implementation of the API, so this commit adds the limit and returns only active questions.

We also only want to send back active conversations, so i've used the active scope in our `find_conversation` method.

### DIFF
--- a/app/models/conversation.rb
+++ b/app/models/conversation.rb
@@ -13,10 +13,11 @@ class Conversation < ApplicationRecord
        },
        prefix: true
 
-  def questions_for_showing_conversation
-    Question.where(conversation: self)
-            .includes(answer: %i[feedback sources])
-            .active
-            .last(Rails.configuration.conversations.max_question_count)
+  def questions_for_showing_conversation(only_answered: false)
+    scope = Question.where(conversation: self)
+                  .includes(answer: %i[feedback sources])
+                  .active
+    scope = scope.joins(:answer) if only_answered
+    scope.last(Rails.configuration.conversations.max_question_count)
   end
 end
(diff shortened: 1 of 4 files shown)


davidgisbey:
@kevindew thanks i've added that in a second commit

app/models/conversation.rb
```diff
@@ -19,4 +19,12 @@ def questions_for_showing_conversation
             .active
             .last(Rails.configuration.conversations.max_question_count)
   end
+
+  def answered_questions_for_api_conversation
```
kevindew, line 23: As this is a very similar to the method above I wonder if we want to just add a parameter on it of answered_only: true|false ?

I did find myself staring for a while to try work out why we've got the two methods that were the same until I realised about the answer joins.

app/models/conversation.rb
```diff
@@ -13,10 +13,11 @@ class Conversation < ApplicationRecord
        },
        prefix: true
 
-  def questions_for_showing_conversation
-    Question.where(conversation: self)
-            .includes(answer: %i[feedback sources])
-            .active
-            .last(Rails.configuration.conversations.max_question_count)
+  def questions_for_showing_conversation(answered: false)
```
kevindew, line 16: Nitpick I think only_answered would work better as with answered: false you still get answered in there.

kevindew:
You've got a couple of commits to rebase here, what's a `:wop` ? 😁

kevindew (approved):

PR #276, opened 2025-06-05
Remove Early Access User Management from the Admin Area

Sunny-xyz:
Trello: https://trello.com/c/zhI0UIU3/2464-remove-earlyaccessuser-and-waitinglistuser-functionality-from-the-admin-area

WIP


Sunny-xyz:
Before:
image

After:
image

spec/lib/slack_poster_spec.rb
```diff
@@ -30,17 +30,6 @@
...
-        user = create(:early_access_user)
-
-        expect(slack_poster).to receive(:send_message).with(
-          "A new user has been shadow banned. <http://chat.dev.gov.uk/admin/early-access-users/#{user.id}|View user>",
-        )
-        described_class.shadow_ban_notification(user.id)
-      end
-    end
-
     describe ".waiting_list_full" do
```
davidgisbey, line 33: we don't need to do this as part of this PR, but it looks like we need to strip this out too. Looks like it was (understandably) missed when we pulled the sign up flow out

PR #278, opened 2025-06-05
Remove Notify integration

davidgisbey:
## Description 

Now we're no longer sending any emails to users we can remove our Notify integration and ActionMailer. One thing that's probably worth noting is i've tried to retain the default action mailer config, but just commented it out.

I.e. what would happen if you ran rails app:update. That way if we ever go back to using ActionMailer we can just uncomment those parts. I've removed all the Notify related config though.

## Trello card

https://trello.com/c/RxzQTX8Z/2534-remove-notify-integration


jackbot (approved):
Nice one, good job on reverting everything to default :+1:

Shall we remove [this env var](https://github.com/alphagov/govuk-chat/blob/main/app.json#L34-L36) for Heroku, as it no longer does anything too? No need for a re-review if you think so.

PR #279, opened 2025-06-05
Remove Passwordless 

davidgisbey:
## Description

Now we've stripped out EarlyAccessUser functionality and the signup/in flow, we can remove the passwordless dependency and all related code since we no longer need to send magic links to users.

## Trello card

https://trello.com/c/D7jjOwHv/2553-remove-passwordless-from-chat


chaecramb (approved):
LGTM

PR #281, opened 2025-06-09
Output answer strategy and embedding provider during evaluation rake task

davidgisbey:
## Description

As the answer strategy and embedding strategy are controlled by env variables it is easy to run the task expecting one thing and getting another. It is also quite a slow task so an early indication is good. Finally having the data of what it was when looking at the data is helpful for clarifying what was used to generate the data.

While it’s a bit of a shame the same data has to be repeated to every row, we don’t really have another option with the JSONL format.

## Trello card

https://trello.com/c/fFWFHfUC/2540-output-information-about-strategy-and-embedding-provider-for-the-evaluationgeneratereport-task


chaecramb (approved):
Looks good

PR #282, opened 2025-06-09
Remove early access models

chaecramb:
This PR removes all database tables, ActiveRecord models, concerns, helpers, and related code for the legacy pilot user system (EarlyAccessUser, WaitingListUser, and their deleted counterparts) in order to clean up the legacy user system from the codebase and database after all dependent functionalities in the chat and admin interfaces have been removed


db/migrate/20250609113203_remove_early_access_user_id_from_conversations.rb
```diff
@@ -0,0 +1,11 @@
+class RemoveEarlyAccessUserIdFromConversations < ActiveRecord::Migration[8.0]
+  def up
+    remove_index :conversations, :early_access_user_id
```
davidgisbey, line 3: bit of a nit but removing a column automatically removes the index

db/migrate/20250609115100_remove_pilot_user_tables.rb
```diff
@@ -0,0 +1,47 @@
+class RemovePilotUserTables < ActiveRecord::Migration[8.0]
+  def up
+    drop_table :early_access_users
+    drop_table :waiting_list_users
+    drop_table :deleted_early_access_users
+    drop_table :deleted_waiting_list_users
+  end
+
+  def down
+    create_enum :waiting_list_users_source, %w[admin_added insufficient_instant_places]
```
davidgisbey, line 10: do these enums not still exist in the db? I can see you've only dropped tables above and not the enums? I think if you used `rails db:rollback` this would blow up due to that

davidgisbey (commented):
looks good. Just a couple of minor comments on a couple of the migrations

db/migrate/20250609115100_remove_pilot_user_tables.rb
chaecramb: Ah yeah, good catch, didn't twig that I'd left them here too.

davidgisbey (approved):
Thanks for making those changes. Looks great 👍

PR #286, opened 2025-06-10
API: return 503 when access is disabled

jackbot:
This uses the `Settings` model to figure out if the API is enabled or
not. If it's disabled, we return a 503 status code and no body.


app/controllers/api/v0/conversations_controller.rb
```diff
@@ -1,4 +1,4 @@
-class Api::V0::ConversationsController < ApplicationController
+class Api::V0::ConversationsController < Api::BaseController
   skip_before_action :verify_authenticity_token
```
davidgisbey, line 2: I think this can probs go in the base controller too

davidgisbey (approved):
left 1 minor comment but not a blocker

app/controllers/api/v0/conversations_controller.rb
jackbot: Great shout, thanks!

PR #288, opened 2025-06-11
Associate web conversation with Signon user

davidgisbey:
## Description

We associate API conversations with a signon user to ensure we track who initiated the conversation. This updates the web version to do the same for consistency.

It also updates the show question page to show the Signon user for conversations created via the web or API and updates the text from "API user" to "Signon user"

## Screenshot

image

## Trello card

https://trello.com/c/5hAzumge/2530-store-when-signon-users-interact-with-web-interface

### DIFF
--- a/spec/requests/conversations_spec.rb
+++ b/spec/requests/conversations_spec.rb
@@ -1,5 +1,8 @@
 RSpec.describe "ConversationsController" do
   delegate :helpers, to: ConversationsController
+  let(:signon_user) { create(:signon_user) }
+
+  before { login_as(signon_user) }
 
   it_behaves_like "handles a request for a user who hasn't completed onboarding",
                   routes: { show_conversation_path: %i[get], update_conversation_path: %i[post] }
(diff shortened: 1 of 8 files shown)


spec/requests/conversations_spec.rb
```diff
@@ -253,6 +256,13 @@
           .to have_selector("h1", text: "GOV.UK Chat is generating an answer")
       end
 
+      it "associates the signon user with the conversation" do
+        post update_conversation_path, params: { create_question: { user_question: "How much tax should I be paying?" } }
+
+        conversation = Conversation.includes(:signon_user).last
+        expect(conversation.signon_user).to eq(signon_user)
```
jackbot, line 263: Is this test a little redundant as you create the conversation in the `let` block with the `signon_user` association, and then verify that the association is still there. Or am I missing something in how this works?

jackbot: Nice one, thanks for explaining.

PR #289, opened 2025-06-11
Remove unused Heroku early access env var

davidgisbey:
## Description

While working on something else i noticed that this isn't being used for anything anymore so we can remove it.


jackbot (approved):
Nice one

PR #290, opened 2025-06-11
Remove downtime_type for public access

jackbot:
https://trello.com/c/pnENCZ5f/2529

We don't want to differentiate between downtime types for public access any more. The public facing app is either on or off, like the API access.

This removes:

* The settings column and enum from the db
* The settings enum from the model
* All admin view logic for downtime type
* The logic for rendering the status code when public access is disabled; we now only use 503


davidgisbey (approved):
Thanks for making those changes. Looks good 👍

PR #295, opened 2025-06-12
Rename public_access to web_access

jackbot:
Now that we have `api_access_enabled`, it makes sense to rename `public_access` to `web_access`.

This removes the column from the settings table, and adds the new one. It then removes all references to `public_access` and replaces them with `web_access`.

- **Add web_access_enabled to Settings**
- **Rename public_access_enabled to web_access_enabled**
- **Rename PublicAccessForm to WebAccessForm**
- **Use web_access_enabled in Settings spec**
- **Admin panel: public_access to web_access**
- **Remove public_access_enabled from Settings**


davidgisbey (approved):
looks good.

PR #298, opened 2025-06-12
Allow overriding of Claude Bedrock config via env vars

jackbot:
https://trello.com/c/qMYt0Bvy/2543

Claude Sonnet 4 is currently only available in US regions, which means
we need to set the AWS region to `us-east-1` in order to use it.

But we don't need to change the region for testing other Claude
versions, or any other Bedrock API calls.

This commit allows both the `CLAUDE_AWS_REGION` and
`CLAUDE_SONNET_MODEL_ID` env vars to be set so that you can test using
Claude Sonnet 4 by using:

```
CLAUDE_AWS_REGION=us-east-1 CLAUDE_SONNET_MODEL_ID=us.anthropic.claude-sonnet-4-20250514-v1:0 <cmd>
```

### DIFF
--- a/lib/bedrock_models.rb
+++ b/lib/bedrock_models.rb
@@ -1,4 +1,4 @@
 module BedrockModels
-  CLAUDE_3_7_SONNET = "eu.anthropic.claude-3-7-sonnet-20250219-v1:0".freeze
+  CLAUDE_SONNET = ENV.fetch("CLAUDE_SONNET_MODEL_ID", "eu.anthropic.claude-3-7-sonnet-20250219-v1:0").freeze
   TITAN_EMBED_V2 = "amazon.titan-embed-text-v2:0".freeze
 end
--- a/lib/answer_composition/pipeline/claude/question_rephraser.rb
+++ b/lib/answer_composition/pipeline/claude/question_rephraser.rb
@@ -11,7 +11,7 @@ module AnswerComposition::Pipeline
       def call
         response = bedrock_client.converse(
           system: [{ text: config[:system_prompt] }],
-          model_id: BedrockModels::CLAUDE_3_7_SONNET, # TODO: change this to a more basic model
+          model_id: BedrockModels::CLAUDE_SONNET, # TODO: change this to a more basic model
           messages:,
           inference_config:,
         )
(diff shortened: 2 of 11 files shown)


lib/bedrock_models.rb
```diff
@@ -1,4 +1,4 @@
 module BedrockModels
-  CLAUDE_3_7_SONNET = "eu.anthropic.claude-3-7-sonnet-20250219-v1:0".freeze
+  CLAUDE_SONNET = ENV.fetch("CLAUDE_SONNET_MODEL_ID", "eu.anthropic.claude-3-7-sonnet-20250219-v1:0").freeze
```
jackbot, line 2: I tried adding a spec for this but because it's a frozen const, it's tricky to test both the default value and the overridden value.

When RSpec starts, it loads the whole Rails app so it'll define this constant to be the default value. Using CodeClimate to set the `CLAUDE_SONNET_MODEL_ID` env var in a single test does nothing because the constant has already been defined before the overridden env var is set.

The test I came up with looked like this:

```ruby
RSpec.describe BedrockModels do
  describe "CLAUDE_SONNET" do
    it "returns the default value" do
      if described_class.const_defined?(:CLAUDE_SONNET)
        described_class.send(:remove_const, :CLAUDE_SONNET) # rubocop:disable RSpec/RemoveConst
      end
      load Rails.root.join("lib/bedrock_models.rb")

      expect(described_class::CLAUDE_SONNET).to(eq("eu.anthropic.claude-3-7-sonnet-20250219-v1:0"))
    end

    it "returns the overridden value set by the env var" do
      ClimateControl.modify(CLAUDE_SONNET_MODEL_ID: "my-model") do
        if described_class.const_defined?(:CLAUDE_SONNET)
          described_class.send(:remove_const, :CLAUDE_SONNET) # rubocop:disable RSpec/RemoveConst
        end
        load Rails.root.join("lib/bedrock_models.rb")

        expect(described_class::CLAUDE_SONNET).to eq("my-model")
      end
    end
  end
end
```

So it removes the constant and reloads the file so the constant gets redefined. It does work, but it feels pretty gross. Not only do you have to go against Rubocop, but it also outputs a bunch of warnings in the RSpec output:

```
warning: already initialized constant BedrockModels::TITAN_EMBED_V2
/govuk/govuk-chat/lib/bedrock_models.rb:3: 
warning: previous definition of TITAN_EMBED_V2 was here
```

Because when we load the `lib/bedrock_models.rb` file, it redefines the `TITAN_EMBED_V2` constant and warns about that. This will only get worse the more constants we define in this file.

So I decided to just not test this particular line. It feels like innocuous enough that I was happier do to that than have this somewhat hacky test.

davidgisbey: Yea i've run into this before too. I know we often pop stuff in the application config and stub that, but at that point you're just testing that you stubbed it not that the env var set the correct model value anyway so not much point.

davidgisbey (approved):
looks good!

PR #302, opened 2025-06-16
BigQuery: export conversation source and signon user

jackbot:
https://trello.com/c/yp95vKmF/2533

Exports the conversation source and signon user to BigQuery. If the
conversation doesn't have a signon user, it will be set to nil.


davidgisbey (approved):
looks good. I've left 1 minor comment about cleaning up an unnecessary pre-existing db hit, but not a blocker.

PR #304, opened 2025-06-16
Migrate from AWS Bedrock SDK to Anthropic's SDK

davidgisbey:
## Description 

This ports us over from using the Bedrock runtime SDK provided by Amazon to using Anthropics SDK. We've run into a few issues with AWS's abstraction, most recently with prompt caching which is discussed in detail below.

This PR:

1. Adds some stubs for hitting the invoke endpoint using Antrhopics SDK
2. Adds a tag which stubs out all credential related requests. This means we only make 1 call to the invoke endpoint
3. Updates each pipeline step to use Anthropics SDK. I've done 1 per commit for ease of review
4. temporarily points the private repo branch to one that splits the contexts out into a second prompt. It also makes a couple of changes to the output schemas so they adhere to the shape anthropic expects (basically removing the json key that has the schema nested within it)

## Prompt caching

Bedrock supports prompt caching for Claude LLM requests. See https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html for more details.

In order for the a prompt to be cached, it needs to meet a few conditions.

1. The prompt must be long enough to exceed the minimum token threshold (1024 tokens for Claude 3.7 Sonnet).
5. The prompt must be an exact match of a previous cache. This means that even a single whitespace difference will prevent it from being cached.

Porting to anthropic allows us to cache individual system prompt messages. The Bedrock SDK didn't allow this and only allowed the caching of all system messages messages cumulatively. 

This was one of the reasons we ported over to using Anthropics SDK.

Due the the above constraints i've added system prompt caching to:

1. Question routing
2. Structured answer composition
3. Guardrail checks

Each of these has static content that exceeds the minimum token threshold and can be cached.

We won't be able to cache the system prompts for the Jailbreak or Question rephraser calls as there's nowhere close to 1000 tokens of static content for either of them. The instructions we send over are concise (~200 tokens for rephraser and ~450 for jailbreak).

## Trello card

https://trello.com/c/lMffS18A/2528-investigate-bedrock-token-cache


lib/answer_composition/pipeline/claude/structured_answer_composer.rb
```diff
@@ -10,29 +10,32 @@ def initialize(context)
...
         start_time = Clock.monotonic_time
-
-        response = bedrock_client.converse(
-          system: [{ text: system_prompt }],
-          model_id: BedrockModels::CLAUDE_SONNET,
+        response = anthropic_bedrock_client.messages.create(
+          system: [
+            { type: "text", text: system_prompt, cache_control: { type: "ephemeral" } },
+            { type: "text", text: context_prompt },
+          ],
```
kevindew, line 17: I wonder as these are both types of system prompt whether we'd be more explicit if they were both named system_prompt, maybe cached_system_prompt, context_system_prompt

lib/answer_composition/pipeline_runner.rb
```diff
@@ -30,7 +30,7 @@ def call
         status: "error_answer_service_error",
         error_message: error_message(e),
       )
-    rescue Aws::Errors::ServiceError => e
+    rescue Anthropic::Errors::APIError => e
```
kevindew, line 33: I imagine we might want a follow up story soon to look into what errors we can get more for logging purposes

spec/lib/answer_composition/pipeline/claude/structured_answer_composer_spec.rb
```diff
@@ -38,102 +33,102 @@
...
-        tool_name: "output_schema",
-      )
-
-      stub_bedrock_converse(response)
+      stub_anthropic_claude_structured_answer_response(question.message, "answer")
 
       described_class.call(context)
-      expect(context.answer.llm_responses["structured_answer"]).to match(response)
+
+      expected_llm_response = {
```
kevindew, line 40: This looks a bit verbose, perhaps there's something else we can do?

kevindew (commented):
Super job here @davidgisbey - I've added a few comments, some on the testing are a bit foundational so might be good to chat about them.

I have found it a bit of a challenge to review so I do expect I might want to look into a bit more on the next review and have comments I missed. In particular I really haven't given the individual tests much attention.

I'm guessing at this stage there isn't really anything practical we can do to split this up so we may have to put together a list of things to look at after it so this doesn't block things as it is pretty beasty.

Gemfile
```diff
@@ -20,7 +21,7 @@ gem "gds-api-adapters"
 gem "gds-sso"
 gem "google-cloud-bigquery", require: false
 gem "govuk_app_config"
-gem "govuk_chat_private", github: "alphagov/govuk_chat_private"
+gem "govuk_chat_private", github: "alphagov/govuk_chat_private", branch: "add-context-prompt-to-claude-structured-answer"
```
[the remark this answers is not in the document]
davidgisbey: I can definitely split the system prompt out into two separate ones, but i won't be able to tweak the output schema structure until the port. I'll sort that first though.

lib/answer_composition/pipeline/claude/question_rephraser.rb
```diff
@@ -27,16 +27,16 @@ def call
 
       attr_reader :question_message, :message_records
 
-      def bedrock_client
-        @bedrock_client ||= Aws::BedrockRuntime::Client.new(
-          region: ENV.fetch("CLAUDE_AWS_REGION", "eu-west-1"),
+      def anthropic_bedrock_client
+        @anthropic_bedrock_client ||= Anthropic::BedrockClient.new(
+          aws_region: ENV.fetch("CLAUDE_AWS_REGION", "eu-west-1"),
```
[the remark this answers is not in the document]
davidgisbey: Okay cool maybe we can add the wrapper and track rate limits as a bit of follow up work?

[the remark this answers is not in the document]
kevindew: Oh yes definitely follow up for wrapper/rate limits - didn't mean to imply in scope for this.

Gemfile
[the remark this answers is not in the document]
davidgisbey: yea great shout i'll do that now

lib/answer_composition/pipeline/claude/question_rephraser.rb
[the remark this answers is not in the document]
davidgisbey: So if i make this change i get failures in CI.

```
  Failure/Error:
        @anthropic_bedrock_client ||= Anthropic::BedrockClient.new(
          aws_region: ENV["CLAUDE_AWS_REGION"],
        )

      Aws::Errors::MissingRegionError:
        No region was provided. Configure the `:region` option or export the region name to ENV['AWS_REGION']
```
We largely just use this for testing in rake tasks right? So i think we will either need to leave it as is. Or remove it. In which case it defaults to `eu-west-1`

spec/support/stub_claude_messages.rb
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    if cached
+      matchers[:system] = array_including(
+        a_hash_including(
+          "type" => "text",
+          "cache_control" => { "type" => "ephemeral" },
+        ),
+      )
+    end
```
kevindew, line 30: I think we've got something quite specific here at quite a low level. I'd probably suggest we treat system the same as we do for messages and treat a string or a regex as a match array including otherwise use it verbatim. 

Then could have a claude_messages_cached_system_prompt helper that returned the array including aspect of this.

spec/support/stub_claude_messages.rb
```diff
@@ -0,0 +1,165 @@
...
+  def claude_messages_tool_use_block(input:, name:, id: "tool-use-id")
+    Anthropic::Models::ToolUseBlock.new(
+      id:,
+      name:,
+      input:,
+      type: :tool_use,
+    )
+  end
+
+  def claude_messages_text_block(text:)
```
kevindew, line 139: super duper minor but a regular argument than a keyword one seems a bit more suited for this.

lib/answer_composition/pipeline_runner.rb
kevindew: Great - I imagine as a follow up for this work we probably want to see what happens if we send too many tokens - as per where we have the ContextLengthExceeded OpenAI error - I guess with the larger context windows it might be so rare that it's not worth logging specifically and thus may need some code changes to remove the error state.

kevindew: Also quick note that as a follow up we may want to see what happens when a tool call exceeds tokens allowed.

lib/answer_composition/pipeline/claude/question_rephraser.rb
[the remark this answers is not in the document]
davidgisbey: Yea, i think i know what's going on actually. It's assuming it should be set in the aws config since the tokens are present. If i add it back in i.e.

```
Aws.config.update(
  credentials: Aws::Credentials.new("fake_access_key", "fake_secret_key"), 
  region: "eu-west-1",
)
```

it works as expected

spec/support/stub_claude_messages.rb
davidgisbey: So i think i'll pull this out of the PR and add it in a follow up one if that's okay just because this PR is already pretty chunky.

Just to clarify. Are you expecting this to be passed into the `stub_claude_messages_response` too? Most of the time we don't care so i'm wondering if it should be optional if that's the case?

davidgisbey: Actually i ended up doing it as part of this PR

lib/answer_composition/pipeline_runner.rb
davidgisbey: I also just looked into what is returned when you send too many tokens. You get a 400 

```
{url: "https://bedrock-runtime.eu-west-1.amazonaws.com/model/eu.anthropic.claude-sonnet-4-20250514-v1%3A0/invoke", status: 400, body: {message: "Input is too long for requested model."}} (Anthropic::Errors::BadRequestError)
```

kevindew: Super thanks for checking both of those. They sound good. 

I think with the latter information we should go ahead and remove the context exceeded error from the system since it seems only easy to detect with OpenAI and we've never experienced it. I've put in a placeholder card: https://trello.com/c/Q428F5Og/2585-remove-the-context-length-exceeded-error-state

davidgisbey: hmm when i copy pasted those messages the content has disappeared visually despite it being in the edit textarea

image

Quite a weird bug! But i've moved the content to the top level comment for visibility.

PR #309, opened 2025-06-17
Retrieval eval: return score

jackbot:
We already return the weighted score in this task, but DS want the
original score to be returned as well. This value is the raw score we
get back from OpenSearch about the search result.

See comment here[1].

[1]: https://github.com/alphagov/govuk-chat-evaluation/pull/49#discussion_r2150308373


kevindew (approved):
Looks good. Only thing I pondered is whether we wanted to prefix the score key with something (opensearch_score, original_score?) as having a `weighted_score` and `score` adds ambiguity. Not too worried if this isn't a concern though, I notice we also have this same issue on the result object.

PR #313, opened 2025-06-18
Conversation questions pagination spike

jackbot:
https://trello.com/c/vxAFbLGm/2537

This changes the `/api/v0/conversation/:id` endpoint to only return the most recent 50 questions.

It then adds a new `/api/v0/conversation/:id/questions` endpoint which returns pages of 50 answered questions for that conversation. This endpoint optionally takes 2 new parameters: `before` and `after`. These are UUIDs for records and will return the previous/next up to 50 records from the given ID.

The idea is that a client will:

1. Make a request to `/api/v0/conversation/:id`. They get some questions in the response to render. 
2. If the JSON response contains an `earlier_questions_url` field, the client knows there are more questions in the conversation
3. The client makes a request to the `earlier_questions_url` which gives them back the previous 50 questions (or less if there isn't 50)
4. The above response will contain another `earlier_questions_url` field if there are still more questions to load.
5. The client continues requesting pages until there aren't any records left

### DIFF
--- a/app/controllers/api/v0/conversations_controller.rb
+++ b/app/controllers/api/v0/conversations_controller.rb
@@ -97,10 +97,11 @@ class Api::V0::ConversationsController < Api::BaseController
       limit: Rails.configuration.conversations.api_questions_per_page,
     )
 
-    render(
-      json: QuestionBlueprint.render(questions, view: :answered),
-      status: :ok,
-    )
+    json = ConversationQuestions.new(
+      questions: QuestionBlueprint.render_as_hash(questions, view: :answered),
+    ).to_json
+
+    render(json:, status: :ok)
   end
 
 private
--- a/app/controllers/api/v0/conversations_controller.rb
+++ b/app/controllers/api/v0/conversations_controller.rb
@@ -19,14 +19,15 @@ class Api::V0::ConversationsController < Api::BaseController
   end
 
   def show
+    limit = Rails.configuration.conversations.api_questions_per_page
     answered_questions = @conversation.questions_for_showing_conversation(
       only_answered: true,
-      limit: Rails.configuration.conversations.api_questions_per_page,
+      limit:,
     )
     pending_question = @conversation.questions.unanswered.last
     answer_url = pending_question ? answer_path(pending_question) : nil
 
-    earlier_questions_url = if @conversation.active_answered_questions_before?(answered_questions.first&.created_at)
+    earlier_questions_url = if answered_questions.size == limit && @conversation.active_answered_questions_before?(answered_questions.first&.created_at)
                               api_v0_conversation_questions_path(
                                 @conversation, before: answered_questions.first.id
                               )
@@ -97,8 +98,24 @@ class Api::V0::ConversationsController < Api::BaseController
       limit: Rails.configuration.conversations.api_questions_per_page,
     )
 
+    earlier_url = if @conversation.active_answered_questions_before?(questions.first&.created_at)
+                    api_v0_conversation_questions_path(
+                      @conversation,
+                      before: questions.first.id,
+                    )
+                  end
+
+    later_url = if @conversation.active_answered_questions_after?(questions.last&.created_at)
+                  api_v0_conversation_questions_path(
+                    @conversation,
+                    after: questions.last.id,
+                  )
+                end
+
     json = ConversationQuestions.new(
       questions: QuestionBlueprint.render_as_hash(questions, view: :answered),
+      earlier_questions_url: earlier_url,
+      later_questions_url: later_url,
     ).to_json
 
     render(json:, status: :ok)
--- a/app/controllers/api/v0/conversations_controller.rb
+++ b/app/controllers/api/v0/conversations_controller.rb
@@ -1,6 +1,6 @@
 class Api::V0::ConversationsController < Api::BaseController
   before_action { authorise_user!(SignonUser::Permissions::CONVERSATION_API) }
-  before_action :find_conversation, only: %i[show update answer answer_feedback]
+  before_action :find_conversation, only: %i[show update answer answer_feedback questions]
 
   def create
     conversation = Conversation.new(signon_user: current_user, source: :api)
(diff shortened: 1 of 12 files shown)


docs/api_openapi_specification.yml
```diff
@@ -234,6 +232,92 @@ paths:
...
+            format: uuid
+        - name: after
+          in: query
+          required: false
+          schema:
+            type: string
+            format: uuid
+          schema:
+            type: string
+            format: uuid
```
kevindew, line 259: Looks like this was duplicated

docs/api_openapi_specification.yml
kevindew, line 1: I imagine we'll want to append the version of this spec with this change.

spec/models/conversations_spec.rb
```diff
@@ -48,13 +48,70 @@
...
+        after_question = create(:question, conversation:, created_at: 3.hours.ago)
+
+        questions = conversation.reload.questions_for_showing_conversation(
+          after_id: after_question.id,
+        )
+        expect(questions).to eq([expected])
+      end
+    end
+
+    context "when a page_size is provided" do
```
kevindew, line 89: I think perhaps you renamed page_size to limit and not this test?

spec/requests/api/v0/conversations_spec.rb
```diff
@@ -181,6 +198,220 @@
...
+
+        expect(JSON.parse(response.body)["earlier_questions_url"]).to eq(
+          api_v0_show_conversation_path(
+            conversation,
+            before: oldest_question_in_page.id,
+          ),
+        )
+      end
+
+      it "is nil if there are no earlier questions" do
```
kevindew, line 334: nit
```suggestion
      it "has a nil value for earlier_questions_url if there are no earlier questions" do
```

app/blueprints/conversation_questions_blueprint.rb
```diff
@@ -0,0 +1,23 @@
+class ConversationQuestionsBlueprint < Blueprinter::Base
+  association(
+    :questions,
+    blueprint: QuestionBlueprint,
+    view: :answered,
+  ) do |_conversation, options|
```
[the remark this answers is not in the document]
jackbot: I wanted to leave this out but Blueprinter really wants you to pass in an object to the class, regardless of whether you use it or not.

It works OK without an object, until you try to use an `if` option, like we do:

```
field(
  :earlier_questions_url,
  if: ->(_field_name, _conversation, options) { options[:earlier_questions_url].present? },
) do |options|
  options[:earlier_questions_url]
end
```

If you omit `_conversation` from the proc params, it blows up:

```
 Failure/Error: if: ->(_field_name, options) { options[:earlier_questions_url].present? },
     ArgumentError: wrong number of arguments (given 3, expected 2)
```

Because Blueprinter tries to call the proc [with three arguments](https://github.com/procore-oss/blueprinter/blob/main/lib/blueprinter/field.rb#L23).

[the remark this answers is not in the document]
kevindew: Makes sense! Does seem annoying.

Couple of quick counter ideas to just rule out a better approach:

1) Would it better to just have a DataModel: ConversationQuestions = Data.define(:questions, :earlier_questions_url, :later_questions_url) ? (I'd wonder if that actually removes the need for some of the if logic in the blueprint)
2) Is the Blueprint actually helping us out here? Until we have a need to reuse the blueprint maybe just mapping questions as a blueprint and returning the url fields is easy enough at the Controller layer?

I think the former seems more consistent of an approach and can certainly keep what you have now if both seem unappealing

app/controllers/api/v0/conversations_controller.rb
```diff
@@ -19,13 +19,24 @@ def create
...
   def show
-    answered_questions = @conversation.questions_for_showing_conversation(only_answered: true)
+    answered_questions = @conversation.questions_for_showing_conversation(
+      only_answered: true,
+      limit: Rails.configuration.conversations.api_questions_per_page,
+    )
     pending_question = @conversation.questions.unanswered.last
     answer_url = pending_question ? answer_path(pending_question) : nil
+
+    earlier_questions_url = if answered_questions.first&.any_older_questions_in_conversation?
```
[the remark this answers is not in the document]
kevindew: On this endpoint there isn't the option for before and after

docs/api_openapi_specification.yml
jackbot: Good shout. Bumped the patch version.

app/controllers/api/v0/conversations_controller.rb
```diff
@@ -78,6 +89,38 @@ def answer_feedback
...
+  def questions
+    questions = @conversation.questions_for_showing_conversation(
+      only_answered: true,
+      before_id: params[:before].presence,
+      after_id: params[:after].presence,
+      limit: Rails.configuration.conversations.api_questions_per_page,
+    )
+
+    earlier_url = if questions.first&.any_older_questions_in_conversation?
+                    api_v0_show_conversation_path(
```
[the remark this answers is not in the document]
kevindew: Yup another story sounds good. Might need to have a chat about what we think should be covered too it's not quite clear to me what we'd want a flow to be

spec/blueprints/conversation_questions_blueprint_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  let(:conversation) { create(:conversation) }
+
+  describe ".render_as_json" do
+    it "generates the questions JSON" do
+      questions = [
+        create(:question, :with_answer, conversation:),
+        create(:question, :with_answer, conversation:),
+      ]
+
+      loaded_questions = Question.where(id: questions.map(&:id))
```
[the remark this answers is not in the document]
kevindew: Sounds good

docs/api_openapi_specification.yml
```diff
@@ -10,7 +10,7 @@ info:
     Worth noting that onboarding aspects of conversation (Informing user of
     limitations and accepting) are not included as within GOV.UK Chat these
     are a UI construct and not a part of core domain constructs.
-  version: "0.1.0"
+  version: "0.1.1"
```
kevindew, line 13:
```suggestion
  version: "0.2.0"
```

Probably a minor version bump as we've added a feature.

kevindew (approved):
Nice job, just a tiny thing but I don't think that needs another review

PR #322, opened 2025-06-23
Update API to return answer path instead of URL for pending questions

davidgisbey:
## Description 

The URLs returned from blueprints e.g: https://github.com/alphagov/govuk-chat/blob/6403285f34b713b1840d62ac2a4fb5db2cebf38c/app/blueprints/question_blueprint.rb#L14-L18 contain a hardcoded hostname and thus are incorrect if we are using an API gateway.

This PR does a couple of things:

1. updates the QuestionBlueprint to return the path instead of the URL
2. passes the view_context into the conversation and question blueprint instead of calling `Rails.application.routes.url_helpers.api_v0_answer_question_path` within the blueprints
3. configure blueprint specs to have `type: :default` so that we can scope helpers used by them to them in our spec_helper.

While we've updated the blueprints to only use a path. if we ever want to revert to using the full URL, passing in the view_context will mean that URLs are constructed using the request host meaning there won't be any need to munge the url in the app or gateway.

## Trello card

https://trello.com/c/KuJeWB28/2565-fix-host-specific-blueprint-links

### DIFF
--- a/app/blueprints/question_blueprint.rb
+++ b/app/blueprints/question_blueprint.rb
@@ -11,11 +11,10 @@ class QuestionBlueprint < Blueprinter::Base
 
   view :pending do
     field :answer_url do |question|
-      path = Rails.application.routes.url_helpers.api_v0_answer_question_path(
+      Rails.application.routes.url_helpers.api_v0_answer_question_path(
         question.conversation_id,
         question.id,
       )
-      "#{Plek.external_url_for('chat')}#{path}"
     end
   end
 end
--- a/app/blueprints/question_blueprint.rb
+++ b/app/blueprints/question_blueprint.rb
@@ -10,11 +10,8 @@ class QuestionBlueprint < Blueprinter::Base
   end
 
   view :pending do
-    field :answer_url do |question|
-      Rails.application.routes.url_helpers.api_v0_answer_question_path(
-        question.conversation_id,
-        question.id,
-      )
+    field :answer_url do |_question, options|
+      options[:answer_url]
     end
   end
 end
--- a/app/controllers/api/v0/conversations_controller.rb
+++ b/app/controllers/api/v0/conversations_controller.rb
@@ -8,7 +8,11 @@ class Api::V0::ConversationsController < Api::BaseController
 
     if form.valid?
       question = form.submit
-      render json: QuestionBlueprint.render(question, view: :pending), status: :created
+      render json: QuestionBlueprint.render(
+        question,
+        view: :pending,
+        answer_url: answer_path(question),
+      ), status: :created
     else
       render json: ValidationErrorBlueprint.render(errors: form.errors.messages), status: :unprocessable_entity
     end
@@ -17,9 +21,15 @@ class Api::V0::ConversationsController < Api::BaseController
   def show
     answered_questions = @conversation.questions_for_showing_conversation(only_answered: true)
     pending_question = @conversation.questions.unanswered.last
+    answer_url = pending_question ? answer_path(pending_question) : nil
+    options = {
+      answered_questions:,
+      pending_question:,
+      answer_url:,
+    }
 
     render(
-      json: ConversationBlueprint.render(@conversation, answered_questions:, pending_question:),
+      json: ConversationBlueprint.render(@conversation, options),
       status: :ok,
     )
   end
@@ -30,7 +40,11 @@ class Api::V0::ConversationsController < Api::BaseController
     if form.valid?
       question = form.submit
 
-      render json: QuestionBlueprint.render(question, view: :pending), status: :created
+      render json: QuestionBlueprint.render(
+        question,
+        view: :pending,
+        answer_url: answer_path(question),
+      ), status: :created
     else
       render json: ValidationErrorBlueprint.render(
         errors: form.errors.messages,
@@ -79,5 +93,12 @@ private
 
   def question_params
     params.permit(:user_question)
+  end
+
+  def answer_path(question)
+    api_v0_answer_question_path(
+      question.conversation_id,
+      question.id,
+    )
   end
 end
--- a/docs/api_openapi_specification.yml
+++ b/docs/api_openapi_specification.yml
@@ -577,7 +577,7 @@ components:
           type: string
         answer_url:
           description: |
-            A URL that can be polled to check whether the answer is available
+            A relative URL that can be polled to check whether the answer is available
           type: string
           format: uri
         created_at:
--- a/spec/blueprints/question_blueprint_spec.rb
+++ b/spec/blueprints/question_blueprint_spec.rb
@@ -52,7 +52,8 @@ RSpec.describe QuestionBlueprint do
           answer_url:,
         }.as_json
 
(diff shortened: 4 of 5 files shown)


spec/support/view_context_helpers.rb
```diff
@@ -0,0 +1,13 @@
+module ViewContextHelpers
+  def view_context_with_answer_path_double(question)
```
davidgisbey, line 2: i'm aware this isn't particularly extensible. But i think it's highly unlikely we would ever need to extend it to have extra routes stubbed. If we ever do then we can refactor this to something like

```
def view_context_with_answer_path_double(stubbed_routes: {})
   view_context_class = Class.new do
      include Rails.application.routes.url_helpers
    end
   
    instance_double(view_context_class, stubbed_routes)
end 

```

But i think it's overkill given the context
```

File: app/blueprints/question_blueprint.rb [hunk 2]
```diff
@@ -10,12 +10,9 @@ class QuestionBlueprint < Blueprinter::Base
...
   view :pending do
-    field :answer_url do |question|
-      path = Rails.application.routes.url_helpers.api_v0_answer_question_path(
-        question.conversation_id,
-        question.id,
-      )
-      "#{Plek.external_url_for('chat')}#{path}"
+    field :answer_path do |question, options|
+      view_context = options[:view_context]
+      view_context.api_v0_answer_question_path(question.conversation_id, question.id)
```
kevindew, line 15: This looks good 👍 

/cc @jackbot as per our discussion just now about view_context

docs/api_openapi_specification.yml
```diff
@@ -563,7 +563,7 @@ components:
         - id
         - conversation_id
         - message
-        - answer_url
+        - answer_path
```
kevindew, line 566: I wonder if we should still call this a URL - since it is effectively a relative URL.

This ties into @jackbot's work as well where there are URLs for questions these would have a query string so would perhaps be semantically incorrect to call them paths?

I think URL might be the best naming suffix? what do you think? 

(I also feel like this is an unimportant decision I've had to consider a million times before yet still don't have the right answer - feels like there is a missing word for a host and scheme agnostic URL)

spec/support/view_context_helpers.rb
kevindew: Yeah I think this might be a bit heavier than is needed. Let me have a look at what I've seen in past projects to see if there is a lighter alternative. Will hopefully succeed before I call it a day.

kevindew (commented):
Nice one, does seem more elegant with the view_context. I'll just have a look at alternatives for the testing.

spec/support/view_context_helpers.rb
kevindew: I've looked it up and a lightweight version you can use is to create a controller and run view_context. Though we still need to set a request on it so I think we'll still want a helper but have it a little more lightweight. Something like:

```
def test_view_context(request = nil)
  controller = ApplicationController.new
  controller.set_request!(request || ActionDispatch::TestRequest.create)
  controller.view_context
end
```

In theory it could need a specific controller like `Api::V0::ConversationsController` if we did have customisations in the controller used but that seems an edge case to not worry about now.

app/blueprints/question_blueprint.rb
jackbot: A concern I have with this is that it couples the Blueprint directly to `v0` of the API. If we end up building a `v1` version then it's going to be tricky to untangle this (unless I'm missing an obvious solution).

Injecting the URLs means that in the `v0` controller you inject the `v0` URL, and in the `v1` controller you can inject a `v1` URL. 

Do we think this is a valid concern?

docs/api_openapi_specification.yml
jackbot: I see your point here, but to me it makes sense to keep it as `path`. If it were `answer_url` I would, as the consumer, assume that it was a full URL with scheme and host. Plus we're using the Rails `api_v0_answer_question_path` so there's precedence for calling a relative URL a path too.

I can't think of a scenario where you'd see it was called `path` and treat it in such a way that the presence of a query string would cause a problem.

app/blueprints/question_blueprint.rb
kevindew: Yeah that is valid - though perhaps premature as we're not quite sure what v0 vs v1 means. I think our plan is to consider namespacing blueprints once we know what differences are.

Definitely a ❌ against embedding though @davidgisbey do you have concerns about injecting?

davidgisbey: No, no concerns with injecting at all. It looks like this is falling into the camp of, let's do something decent but simple and then revisit it when necessary. Very happy to just inject the url for now if we're all happy with that then.

docs/api_openapi_specification.yml
davidgisbey: yea i went back and forth on this. I went from `answer_url` to `relative_answer_url` to `path`. This was why i ended up using path 

> Plus we're using the Rails api_v0_answer_question_path so there's precedence for calling a relative URL a path too

But maybe it's too Railsy. I think either `answer_url` or `path` is fine. But either way i'll update the docs to make it clear it's a relative URL (which makes me sort of lean towards it being answer_url).

app/controllers/api/v0/conversations_controller.rb
```diff
@@ -17,9 +21,15 @@ def create
   def show
     answered_questions = @conversation.questions_for_showing_conversation(only_answered: true)
     pending_question = @conversation.questions.unanswered.last
+    answer_path = pending_question ? answer_path(pending_question) : nil
+    options = {
+      answered_questions:,
+      pending_question:,
+      answer_path:,
+    }.compact
```
kevindew, line 29: Super minor, but does the compact have consequences?

app/controllers/api/v0/conversations_controller.rb
```diff
@@ -80,4 +94,11 @@ def answer_feedback_params
   def question_params
     params.permit(:user_question)
   end
+
+  def answer_path(question)
+    view_context.api_v0_answer_question_path(
```
kevindew, line 99: I don't think you need to use view_context here I think the controller already has the route methods available

docs/api_openapi_specification.yml
kevindew: Oops somewhere I commented already to this but I think it was lost in the ether.

Yeah the Rails example is a good one, it bugs me a bit because the pedant in me knows that a path with a query string is a valid URL but isn't a path - but I happily live with the Rails helper.

A counter suggestion I had was to use a less Rails loaded term than path or url and go for href or link? what do you think?

davidgisbey: Yea good shout. Particularly because our known consumers aren't using Rails. Makes sense. I have a mild preference for href, but happy to go with link if someone has a strong preference.

davidgisbey: Okay makes sense. Have changed it to answer_url

kevindew (approved):
Nice one

PR #325, opened 2025-06-24
Update env-example to include optional env vars

davidgisbey:
We have some additional environment variables that are useful for local development. Particuarly those related to Amazon Bedrock models.

These can be used to toggle the answer stratgey and embedding provider to use Amazon Bedrock models (Claude Sonnet and Titan) instead of OpenAI.

There's also some env vars for Opensearch. Those are only required if you're not using govuk-docker.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,15 @@
 ## variables for working with OpenAI
 OPENAI_ACCESS_TOKEN=
+
+## Opensearch configuration. Not required for local development using govuk-docker
+OPENSEARCH_URL=
+OPENSEARCH_USERNAME=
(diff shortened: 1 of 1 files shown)


.env.example
```diff
@@ -1,2 +1,15 @@
 ## variables for working with OpenAI
 OPENAI_ACCESS_TOKEN=
```
exfalsoquodlibet, line 2: I wonder, this becomes optional if one sets `ANSWER_STRATEGY=claude_answer_strategy`, right?

davidgisbey: Yes it does. At some point we will want to use Sonnet by default. At which point we can just swap these around.

PR #332, opened 2025-06-27
Explicitly add nil cached tokens for claude question rephraser

davidgisbey:
All other Claude pipeline steps either return the cached number of tokens or nil if we aren't caching the request. This brings the question rephraser in line with the other steps.


kevindew (commented):
Nice one

davidgisbey:
@kevindew i think you meant to approve but didn't

PR #341, opened 2025-07-01
Remove onboarding

kevindew:
Trello: https://trello.com/c/9aPFbtsd/2599-remove-onboarding-flow-from-web-chat

This reduces the onboarding process for starting a conversation with chat over web to a single view with progressively disclosed messages:

Jul-01-2025 15-41-31

To do this I had to remove quite a lot of code and try deal with some reusable modules that have lost some of their purpose. I'm willing to bet I've missed something or left something odd behind - it is your quest, dear reviewer, to find that.

A somewhat janky thing is that if you start a new chat you'll get the onboarding steps again, but as it's quick and the web view is low usage this doesn't seem too big a concern.

---

One thing I spotted was that we had use of a component class outside of a component (naughty) and in this PR I remove the component and thus affect the view's styling (why it's naughty). I've switched to using the [design system button group](https://design-system.service.gov.uk/components/button/secondary-combo/) to try remedy it.

Before:

Screenshot 2025-07-01 at 15 40 24

After:

Screenshot 2025-07-01 at 15 40 15

@JamesCGDS would you mind checking to see if you think this needs extra work?

### DIFF
--- a/app/views/conversations/clear.html.erb
+++ b/app/views/conversations/clear.html.erb
@@ -8,23 +8,25 @@
       option from the menu.
     </p>
 
-    <%= form_with url: clear_path, class: "app-c-onboarding-form" do |f| %>
-      <%= render "components/blue_button", {
-        text: "Start a new chat (clears last chat)",
-        data_attributes: {
-          "module": "ga4-event-tracker",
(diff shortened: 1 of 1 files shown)


jackbot (approved):
Nice one. I've had a test and can't see anything unexpected.

PR #345, opened 2025-07-02
Generate Titan embeddings when synchronising GOV.UK content

jackbot:
https://trello.com/c/3a32SO6I/2604

When we pull content off the GOV.UK queue, we want to generate the Titan
embeddings for it so they can be retrieved as part of the answer
generation process.

We'll leave the OpenAI embedding generation too and remove it at a
later date.

I added the mapping to the production OpenSearch instance yesterday, and that's
now been replicated to integration and staging (and verified it's there on all
envs). When this is merged, I'll put a message on the queue and verify that the
embedding was generated.


kevindew (approved):
Sorry forgot to finish this review, one small comment but shouldn't need another review

PR #349, opened 2025-07-02
Update message component

JamesCGDS:
## What
This PR updates the message component to align it (as far as is practical) with the most recent design work. For additional detail, please see the commit messages.

I've tested in Chrome, Firefox, and Safari. There is a small spacing issue in Safari however I'm thinking of tackling this in another PR as I'm currently not sure how to fix it.

## Why
[Trello card](https://trello.com/c/nbgazXo1/2600-update-message-component)

### Desktop

| Sources component - before | Sources component (collapsed) - after | Sources component (expanded) - after |
|------------------------------|----------------------------------------|----------------------------------------|
| image | image | image |

### Mobile

| Sources component - before | Sources component (collapsed) - after | Sources component (expanded) - after |
|------------------------------|----------------------------------------|----------------------------------------|
| image | image | image |

### DIFF
--- a/spec/system/user_toggles_sources_dropdown_spec.rb
+++ b/spec/system/user_toggles_sources_dropdown_spec.rb
@@ -1,27 +1,23 @@
 RSpec.describe "User toggles the sources dropdown" do
-  scenario "Expanding the dropdown" do
+  scenario do
     given_i_am_a_web_chat_user
     and_i_have_an_active_conversation_with_an_answered_question_with_sources
     when_i_visit_the_conversation_page
-    and_i_click_on_the_sources_dropdown
-    then_the_sources_dropdown_expands
-    and_i_see_the_source_links
-  end
+    then_i_cannot_see_the_source_links
 
-  scenario "Collapsing the dropdown" do
-    given_i_am_a_web_chat_user
-    and_i_have_an_active_conversation_with_an_answered_question_with_sources
-    when_i_visit_the_conversation_page
-    and_i_have_expanded_the_sources_dropdown
+    when_i_click_on_the_sources_dropdown
+    then_i_can_see_the_source_links
+
     when_i_click_on_the_sources_dropdown_again
-    then_the_sources_dropdown_collapses
-    and_i_cannot_see_the_source_links
+    then_i_cannot_see_the_source_links
   end
 
   def and_i_have_an_active_conversation_with_an_answered_question_with_sources
     @conversation = create(:conversation, signon_user: @signon_user)
     set_rack_cookie(:conversation_id, @conversation.id)
-    answer = build(:answer, :with_sources, message: "Example answer")
+    @source_vat = build(:answer_source, exact_path: "/vat", title: "Everything about VAT", relevancy: 0)
+    @source_income_tax = build(:answer_source, exact_path: "/income-tax", title: "Income Tax Details", relevancy: 1)
+    answer = build(:answer, sources: [@source_vat, @source_income_tax], message: "Example answer")
     create(:question, answer:, conversation: @conversation, message: "Example question")
   end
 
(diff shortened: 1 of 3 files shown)


spec/system/user_toggles_sources_dropdown_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    create(:question, answer:, conversation: @conversation, message: "Example question")
+  end
+
+  def when_i_visit_the_conversation_page
+    visit show_conversation_path
+  end
+
+  def and_i_click_on_the_sources_dropdown
+    # click_on seems to work for links and buttons but not the details element, hence the alternative approach below
+    page.find(".app-c-conversation-sources__details").click
```
kevindew, line 34: I'd wonder if we can use the text as something to click on and assert that we can see a source with the expected name rather than using the DOM selector. I'm happy to try have a play with sorting this on Monday as I'm not sure if there are problems to implement it.

kevindew (commented):
From an initial look, this looks good to me but do have a comment on the system test.

I'm also not 100% sure it's valuable to have the system test as it is browser functionality we're testing but naturally we've not got any other tests and it seems consistent with what else we've tested - so I think it's good for now, but perhaps later we'd want to combine it with another more generic test of asking a question and getting answer (particularly as it's not a JS feature)

JamesCGDS:
Thanks for taking a look at this @kevindew. I've made some changes to the system test to target text and directly look for source links (kept it as a separate commit for now - can squash if all looks okay).

Re targeting source links, I wasn't able to reliably test against the link text as it seemed to change. I did try using a regex but this only seemed to work for the `have_content` assertion and not the `have_link` assertion. Therefore, I've ignored the link text in the test and tested against the `href` instead.

Happy to discuss/make any changes (or feel free to change things directly if that's easier).

Also happy to take a look at integrating this into an existing test at some point - I'll make a note of it.

spec/system/user_toggles_sources_dropdown_spec.rb
```diff
@@ -0,0 +1,63 @@
+RSpec.describe "User toggles the sources dropdown" do
+  scenario "Expanding the dropdown" do
```
kevindew, line 2: I'd suggest turning this into a single test to toggle the drop down, with the rational that we can basically test the first one as part of your second test and then can save time

```
    given_i_am_a_web_chat_user
    and_i_have_an_active_conversation_with_an_answered_question_with_sources
    when_i_visit_the_conversation_page
    then_i_cannot_see_the_source_links

    when_i_click_on_the_sources_dropdown
    then_i_can_see_the_source_links

    when_i_click_on_the_sources_dropdown_again
    then_i_cannot_see_the_source_links
```

spec/system/user_toggles_sources_dropdown_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  def and_i_see_the_source_links
+    expect(page).to have_link(nil, href: "https://www.test.gov.uk/income-tax")
+    expect(page).to have_link(nil, href: "https://www.test.gov.uk/vat-tax")
+  end
+
+  def and_i_have_expanded_the_sources_dropdown
+    page.find("details", text: "GOV.UK pages used in this answer (links open in a new tab)").click
+  end
+
+  def when_i_click_on_the_sources_dropdown_again
```
kevindew, line 50: Rather than needing to define this method that is the same as `and_i_click_on_the_sources_dropdown` we can do method aliasing: 

https://github.com/alphagov/govuk-chat/blob/298c554116efb0841255df1018a991b51e69b94b/spec/system/conversation_js_features_spec.rb#L136-L137

spec/system/user_toggles_sources_dropdown_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    and_i_have_expanded_the_sources_dropdown
+    when_i_click_on_the_sources_dropdown_again
+    then_the_sources_dropdown_collapses
+    and_i_cannot_see_the_source_links
+  end
+
+  def and_i_have_an_active_conversation_with_an_answered_question_with_sources
+    @conversation = create(:conversation, signon_user: @signon_user)
+    set_rack_cookie(:conversation_id, @conversation.id)
+    answer = build(:answer, :with_sources, message: "Example answer")
```
kevindew, line 24: To set specific sources we can modify this code to something like:

```
@source_vat = build(:answer_source, exact_path: "/vat", title: "Everything about VAT" relevancy: 0)
@source_income_tax = build(:answer_source, exact_path: "/income-tax", title: "Income Tax Details" relevancy: 1)

answer = build(:answer, sources: [@source_vat, @source_income_tax], message: "Example answer")
```

then we can assert based off the variables too:

```
expect(page).to have_link(@source_vat.title, href: @source_vat.url)
```

kevindew (commented):
Thanks @JamesCGDS your changes look good. I think this just needs a few more refinements then ready to merge.

JamesCGDS:
Thanks for the feedback @kevindew - have implemented those changes now.

PR #357, opened 2025-07-09
Retry Titan embedding generation

jackbot:
We've had an exception come through when generating Titan embeddings
where the input text exceeded the input token limit on Titan [1].

Titan supports either 50,000 characters OR 8,192 tokens for input [2]. We
already truncate to 50,000 characters, but there's no way for us to
know how many input tokens our text would convert to. So it's possible
that the input text is less than 50,000 characters but more than 8,192
tokens.

We used Tiktoken to calculate this for our OpenAI embeddings, but
there's no equivalent for other providers.

Currently an exception will just be raised if the token count is too
high, so this is an attempt to fix that. It will always truncate to
50,000 characters still, then it'll attempt an embedding on the
truncated text. If that fails, we'll try truncating to 80% of the
original text length. We'll continue truncating to 80% for 5 attempts.
If it's still too long then an exception is raised.

These numbers might need tweaking, but it feels like a good place to
start.

The vast majority of our chunks are within the limits, but sometimes a
bit chunk of HTML comes along (e.g. this content item [3]), which pushes it
over the limit.

[1]: https://govuk.sentry.io/issues/6726533431
[2]: https://docs.aws.amazon.com/bedrock/latest/userguide/titan-embedding-models.html
[3]: https://www.gov.uk/api/content/government/publications/balai-directive-approved-premises/list-of-bodies-institutes-and-centres-approved-for-intra-community-trade-in-animals

### DIFF
--- a/lib/search/text_to_embedding/titan.rb
+++ b/lib/search/text_to_embedding/titan.rb
@@ -1,6 +1,9 @@
 class Search::TextToEmbedding
   class Titan
+    class InputTextTokenLimitExceededError < StandardError; end
+
     INPUT_TEXT_LENGTH_LIMIT = 50_000
+    RETRY_COUNT = 5
 
     def self.call(...) = new(...).call
 
@@ -10,13 +13,7 @@ class Search::TextToEmbedding
     end
 
     def call
-      # For strings longer than the embedding model limit we have to truncate
-      # the text.
-      # This is done silently and in future we may want to log this in a
-      # database or log a warning.
-      to_embed = text_collection.map(&method(:keep_input_within_text_limit))
-
-      embeddings = convert_text_to_embeddings(to_embed)
+      embeddings = text_collection.map(&method(:convert_text_to_embeddings))
 
       # return just first embedding rather than an array of embeddings if we
       # weren't given an array input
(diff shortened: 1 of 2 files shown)


lib/search/text_to_embedding/titan.rb
```diff
@@ -31,21 +28,42 @@ def bedrock_client
...
-    def convert_text_to_embeddings(to_embed_collection)
-      to_embed_collection.map do |text|
-        response = bedrock_client.invoke_model(
-          model_id: BedrockModels::TITAN_EMBED_V2,
-          body: {
-            inputText: text,
-          }.to_json,
-        )
+      while tries < RETRY_COUNT
+        text = text[0...(text.length * 0.8)] if tries.positive? # Truncate to 80% of the current length
```
kevindew, line 42: Wonder if it's better to go for 90% then over the course of 5 retries you can lose at most 40% of the length, whereas with 80% you can lose 80% of the length?

The evidence you have with the large string that failed may indicate otherwise though - was that close to the limit or miles over?

kevindew (commented):
Nice work in a tightly constrained situation. Seems a real shame we don't have access to a tokenizer to actually work this out.

lib/search/text_to_embedding/titan.rb
jackbot: Yeah the large string I tested with was:

Original length: 43324 chars

Trying 43324 chars - FAIL
Trying 34659 chars - FAIL
Trying 27727 chars - FAIL
Trying 22181 chars - FAIL
Trying 17744 chars - PASS

So it only passed on the last iteration with 80% truncation each time.

kevindew (approved):
Nice one

PR #358, opened 2025-07-09
Remove the Context length exceeded error state

Sunny-xyz:
Trello: https://trello.com/c/Q428F5Og/2585-remove-the-context-length-exceeded-error-state

This PR removes the dedicated rescue branch for OpenAIClient::ContextLengthExceededError in AnswerComposition::PipelineRunner, allowing those errors to fall through to the existing generic error handler. Claude has context length errors as part of standard 400 errors, and we’ve never seen this error from OpenAI in practice, so unifying it all under one path simplifies the code.

### DIFF
--- a/lib/answer_composition/pipeline_runner.rb
+++ b/lib/answer_composition/pipeline_runner.rb
@@ -16,13 +16,6 @@ module AnswerComposition
       end
 
       context.answer
-    rescue OpenAIClient::ContextLengthExceededError => e
-      GovukError.notify(e)
-      context.abort_pipeline(
-        message: Answer::CannedResponses::CONTEXT_LENGTH_EXCEEDED_RESPONSE,
-        status: "error_context_length_exceeded",
-        error_message: error_message(e),
-      )
     rescue OpenAIClient::RequestError => e
       GovukError.notify(e)
       context.abort_pipeline(
--- a/spec/lib/answer_composition/pipeline_runner_spec.rb
+++ b/spec/lib/answer_composition/pipeline_runner_spec.rb
@@ -47,28 +47,6 @@ RSpec.describe AnswerComposition::PipelineRunner do
       end
     end
 
-    context "when the step raises an OpenAIClient::ContextLengthExceededError" do
-      let(:error) { OpenAIClient::ContextLengthExceededError.new("error message") }
-      let(:pipeline_step) { ->(_context) { raise error } }
-
-      it "notifies sentry" do
-        expect(GovukError).to receive(:notify).with(error)
-        described_class.call(question:, pipeline: [pipeline_step])
-      end
-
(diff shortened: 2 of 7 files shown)


lib/answer_composition/pipeline_runner.rb
```diff
@@ -16,14 +16,7 @@ def call
...
       context.answer
-    rescue OpenAIClient::ContextLengthExceededError => e
-      GovukError.notify(e)
-      context.abort_pipeline(
-        message: Answer::CannedResponses::CONTEXT_LENGTH_EXCEEDED_RESPONSE,
-        status: "error_context_length_exceeded",
-        error_message: error_message(e),
-      )
-    rescue OpenAIClient::RequestError => e
+    rescue OpenAIClient::RequestError, OpenAIClient::ContextLengthExceededError => e
```
kevindew, line 19: So OpenAIClient::ContextLengthExceededError is a subclass of OpenAIClient::RequestError so we don't actually need this to be included in this list as it will match OpenAIClient::RequestError

spec/lib/answer_composition/pipeline_runner_spec.rb
```diff
@@ -62,8 +62,8 @@
           .to be_a(Answer)
           .and have_attributes(
             question:,
-            status: "error_context_length_exceeded",
```
kevindew, line 65: We'll probably want to look at removing this status from being one of our enum of status values (which will be a bit of a pain)

spec/lib/answer_composition/pipeline_runner_spec.rb
```diff
@@ -62,8 +62,8 @@
...
-            message: Answer::CannedResponses::CONTEXT_LENGTH_EXCEEDED_RESPONSE,
```
kevindew, line 66: We won't need this canned response once there isn't code for this error.

kevindew (commented):
Thansk for this, there's a few bits to sort out but looks on the right track 👍

spec/lib/answer_composition/pipeline_runner_spec.rb
```diff
@@ -62,8 +62,8 @@
...
+            status: "error_answer_service_error",
+            message: Answer::CannedResponses::ANSWER_SERVICE_ERROR_RESPONSE,
             error_message: "class: OpenAIClient::ContextLengthExceededError message: error message",
```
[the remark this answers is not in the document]
Sunny-xyz: I'll remove this now

Sunny-xyz:
Thanks a ton for the review Kev! I've cleaned up the PR now and have removed all explicit mentions of the error state. I believe this PR is now ready to be merged. I'll follow your guidance for the second PR that handles the enum value

kevindew (approved):
Great job

PR #363, opened 2025-07-11
Remove openai_embedding from OpenSearch mapping

jackbot:
This is the last step towards switching over to using Titan for our
embeddings.

It removes the `openai_embedding` from the OpenSearch mapping and
removes all references to it from the codebase.

Also switches the default embedding provider to Titan (even though it's
the only supported one now).

There's more clean-up that we could do now that we only have a single
provider (we don't need the env var, application config or subclasses
any more), but it might make sense to keep these here for now in case we
need to support other providers in the future.


kevindew (approved):
Looks good

PR #371, opened 2025-07-14
Update UNSUCCESSFUL_REQUEST_MESSAGE

DavidStanden:
CHANGED
Sorry, something went wrong while trying to answer your question. Try again later.

We saved your conversation. Check [GOV.UK guidance for businesses](https://www.gov.uk/browse/business) if you need information now.

TO
Something went wrong while trying to answer your question. Please try again.

REASON
Updating message for the app private beta.


davidgisbey:
same as here https://github.com/alphagov/govuk-chat/pull/369#issuecomment-3068742181

> I've just made a v minor change. Since it's not a multiline message i've updated it to use a string rather than a heredoc (the <<~MESSAGE that wraps the message).

PR #375, opened 2025-07-14
Remove Unused Answer Status Value

Sunny-xyz:
WIP


db/migrate/20250714112750_remove_unused_answer_status_enum_values.rb
```diff
@@ -0,0 +1,33 @@
+class RemoveUnusedAnswerStatusEnumValues < ActiveRecord::Migration[8.0]
+  def up
+    execute <<-SQL
+    ALTER TYPE answer_status RENAME TO answer_status_old;
+
+    CREATE TYPE answer_status AS ENUM(
+                            'answered',
+                            'banned',
```
kevindew, line 8: Sorry, can I add to the scope of this work and also remove this banned status? you may need to open an additional PR first to remove the config for it. I understand this is also a redundant status but hadn't been removed from this list.

kevindew (commented):
Changes look great, I just noticed something left behind (the banned status) and wanted to see if we could also sort that since the enum migration is a bit big and annoying compared to other migrations.

kevindew (approved):
Nice one

PR #376, opened 2025-07-14
Add ADR for using Anthropic SDK for Claude invocations

davidgisbey:
## Description 

We've recently switched our implementation of Claude model invocations from the AWS SDK for Bedrock to the official Anthropic SDK for Ruby.

This change allows us to leverage the native API semantics provided by Anthropic while still using models hosted on AWS Bedrock.

This ADR documents the decision and rationale for this change.

## Trello card

https://trello.com/c/8ni9SEEm/2624-adr-for-anthropic-sdk

### DIFF
--- a/docs/adr/0008-use-anthropic-sdk-for-anthropic-model-invocations.md
+++ b/docs/adr/0008-use-anthropic-sdk-for-anthropic-model-invocations.md
@@ -1,0 +1,32 @@
+# 8. Use Anthropic SDK for Anthropic models
+
+**Date:** 2025-07-14
+
+## Context
+
+We are using Claude models for answer composition in GOV.UK Chat. Initially, we chose to integrate with Claude via AWS Bedrock’s `converse` API using the `aws-sdk-bedrockruntime` gem. At the time, this offered a unified interface for accessing multiple model providers and avoided the lower-level concerns of working with `invokeModel` directly—such as having to manually parse payloads.
+
+However, over time, we ran into several limitations with this approach:
+
+- We often had to consult both AWS and Anthropic documentation to understand the API structure and behavior, since the Converse API is abstracted to work with many models and does not fully mirror Anthropic's interface.
+- Certain Claude features, like `disable_parallel_tool_use`, were either unsupported or undocumented by the Bedrock SDK.
+- Prompt caching was difficult to manage. While Bedrock technically supports `cache_control`, it only supports caching the entire system prompt. Splitting static and dynamic sections into separate system messages did not work reliably—the cache would only match if the same set of chunks appeared in the exact order, making partial caching effectively unusable.
+
+Since our original implementation, Anthropic released an official Ruby SDK that includes native support for calling Claude models hosted on AWS via `Anthropic::BedrockClient`. This presents a cleaner and more predictable implementation and is aligned with Anthropic’s own API documentation.
+
+## Decision
+
+We have ported our Claude integration from the AWS SDK for Bedrock to the [official Anthropic SDK for Ruby](https://github.com/anthropics/anthropic-sdk-ruby), specifically using its `Anthropic::BedrockClient` interface. This allows us to continue using Claude models hosted on AWS Bedrock, but through Anthropics SDK which is designed to expose the full range of Claude's features.
+
+## Status
+
+**accepted**
+
+## Consequences
+
+- All Claude interactions now use the Anthropic SDK rather than the AWS Bedrock Converse API
+- We gain reliable, partial system prompt caching using `cache_control`
+- Claude features not available via the Converse API—such as citations are now accessible
+- Testing integrations is easier, particularly across platforms (e.g., Bedrock and GCP Vertex)
+- If we need to access features that are only exposed via Converse (and not supported by the Anthropic SDK), we may need to revert back to using the Converse API
+- We continue to use the AWS Bedrock SDK directly for generating embeddings


docs/adr/0008-use-anthropic-sdk-for-anthropic-model-invocations.md
```diff
@@ -0,0 +1,33 @@
...
+
+**Date:** 2025-07-14
+
+## Context
+
+We are using Claude models for answer composition in GOV.UK Chat. Originally, we integrated our Claude invocations via AWS Bedrock’s `converse` API using the `aws-sdk-bedrockruntime` gem. While this gave us access to Claude models we host on AWS Bedrock, it introduced some limitations, particularly around system message caching.
+
+We want the system instructions in the system prompt to be cached, but the dynamic portion of the prompt (the context, which varies per request) not to be. OpenAI provides this functionality by default, even when a single system message includes both, but this proved impossible using the Bedrock SDK.
+
+Bedrock does support `cache_control`, but it only allows the caching of the whole system prompt. Attempts to add two system messages—one for the static and the other for dynamic content—were ignored. Each time, it would write to the cache but wouldn't match unless the exact set of chunks were passed to the context in the exact same order.
```
kevindew, line 11: There's probably a bit more context to include than the caching side of things. 

Summarised I think it'd be a long the lines of:

- We began using Claude using the Bedrock converse API as this seemed advantageous if we wished to switch models and better experience to using invokeModel which required some low level concerns, such as JSON parsing
- We found however that working an abstraction away from the underlying Anthropic could be problematic as we had to consult both Anthropic and AWS documentation to understand API structure
- We struggled to understand behaviour that wasn't explicitly documented in AWS API such as `disable_parallel_tool_use` 
- We wanted to cache some of our prompts but not which is a feature unavailable on Bedrock Converse API
- In the time since our initial developer there is now an Anthropic Ruby API with AWS Bedrock support, which provides a nicer user experience than using the AWS invokeModel endpoint

docs/adr/0008-use-anthropic-sdk-for-anthropic-model-invocations.md
```diff
@@ -0,0 +1,33 @@
...
+
+We have chosen to port our Claude integration from the AWS SDK for Bedrock to the [official Anthropic SDK for Ruby](https://github.com/anthropics/anthropic-sdk-ruby), specifically using its built-in `Anthropic::BedrockClient`. This allows us to continue using Claude models hosted on AWS Bedrock, but with more tailored interface provided by the Anthropic-maintained gem.
+
+This change gives us access to Anthropic’s native API semantics while accessing models hosted om AWS Bedrock. In particular, it resolves issues we encountered with prompt caching.
+
+### Rationale
+
+Using the Anthropic SDK’s `BedrockClient` gives us reliable support for `cache_control` at the individual message level, allowing us to cache only the static portion of the system prompt. Beyond caching, the SDK is more closely aligned with Anthropic’s native interface. This improves clarity, reduces surprises, and helps ensure our implementation stays consistent with how Claude is intended to be used.
+
+We will continue to use `aws-sdk-bedrockruntime` for other parts of our LLM pipeline that use non-Anthropic models hosted on AWS. For example, when using Titan to generate embeddings.
```
kevindew, line 23: I'm not sure you actually need this section since we kinda cover it with the context and decision.

docs/adr/0008-use-anthropic-sdk-for-anthropic-model-invocations.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Status
+
+**accepted**
+
+## Consequences
+
+- All Claude interactions now use the Anthropic SDK
+- We now have reliable control over system message caching using `cache_control`
+- We continue to use the Bedrock Converse API for Titan embeddings, as the Anthropic SDK does not provide embedding support
```
kevindew, line 33: We've got a few more consequences I'd imagine:

- We've got access to Claude features that are not supported by Converse API such as citations
- To use any Converse API specific features not supported by Anthropic SDK we will need to switch back
- It is easier to test Anthropic integrations, such as first party Anthropic and GCP Vertex (aside: this was very useful for me to test latency 🥳 )
- We won't benefit from the common interface to test other models

kevindew (commented):
Thanks for this David, it looks great but I'd like to broaden it to refer to more of our pains than just the caching one which I see as being the straw that broken the camel's back

docs/adr/0008-use-anthropic-sdk-for-anthropic-model-invocations.md
```diff
@@ -0,0 +1,32 @@
...
+**accepted**
+
+## Consequences
+
+- All Claude interactions now use the Anthropic SDK rather than the AWS Bedrock Converse API
+- We gain reliable, partial system prompt caching using `cache_control`
+- Claude features not available via the Converse API—such as citations are now accessible
+- Testing integrations is easier, particularly across platforms (e.g., Bedrock and GCP Vertex)
+- If we need to access features that are only exposed via Converse (and not supported by the Anthropic SDK), we may need to revert back to using the Converse API
+- We continue using the Bedrock Converse API for non-Anthropic Bedrock hosted models such as Titan for embeddings
```
kevindew, line 32: This last point isn't correct as we use [InvokeModel for embeddings](https://github.com/alphagov/govuk-chat/blob/2af3ff85575096afe25e26e956102858e63c6cf3/lib/search/text_to_embedding/titan.rb#L64-L69). It's probably more accurate to say we continue to use the AWS Bedrock SDK directly for generating embeddings.

kevindew (approved):
Looks great, one thing to fix but I don't think it needs another review

PR #377, opened 2025-07-14
Remove Banned Status from answer status enum and config

Sunny-xyz:
This PR drops the now‐unused banned value from our answer_status configuration. We’ve already rebuilt the enum type itself to exclude 'banned' as part of PR #375   , so this change cleans up the corresponding YAML and model definitions.


kevindew (approved):
Super, thanks for this

PR #379, opened 2025-07-15
Remove feedback from the API

jackbot:
https://trello.com/c/jHcOtdeA/2623

The mobile app doesn't make use of feedback, so there's no need to keep it in the API.


davidgisbey (approved):
looks good ✅

PR #388, opened 2025-07-17
Bump govuk_chat_private and add new question routing tool config

davidgisbey:
## Description

This adds two enum values to the question_routing_label enum. `unclear_intent` and `requires_account_data`. It adds the required config to questioning routing label config in [config/question_routing_labels.yml](https://github.com/alphagov/govuk-chat/pull/388/commits/222a1cb27de92f0b010ebe22412d09249de45778#diff-c505b93426c982f194635ea1dbfe541a79c8d17153485f6a02630182526f4dd0).

Finally it bumps the max tokens to 500 in the inference config.


davidgisbey:
Great work! Not a big deal at all, but a bit of general feedback on commit messages is we try to keep the summary/title of the commit message to ~50 chars and the commit message to ~75 chars per line so it's a bit more readable/formats better. At the moment, 3 of the 4 commits summaries are quite long 

image

It's really easily resolvable though. We could do something like this for the first commit

```
## current commit message

feat question-routing: updated app/models/answer.rb config/question_r…
…outing_labels.yml db/schema.rb to include new intent tool / question routing label requires_account_data

adding it instead of replacing the deprecated personal_info so that personal_info can still be used in the openai version of Chat; requires_account_data has been only developed for the Claude version so far


## New commit message 

Add requires_account_data question_routing_label

This adds a new requires_account_data question_routing_label enum in the answer model and adds configuration to the question_routing_labels.yml file.

This enum replaces the personal_info for Claude answer generation, but personal_info is still used in the OpenAI version of chat so we still require that enum value.

```
or similar

db/migrate/20250709105644_add_unclear_intent_and_requires_account_data.rb
```diff
@@ -0,0 +1,6 @@
+class AddUnclearIntentAndRequiresAccountData < ActiveRecord::Migration[8.0]
```
davidgisbey, line 1: I think this commit should come before the commits where the values are used in the model. So 

commit 1 - adds the migration and db schema changes
commit 2 - adds requires_account_data to the enum in the model/adds config etc.
commit 3 - adds unclear_intent to the enum in the model/adds config etc.

davidgisbey:
Looks great 👍 I'm just going to make 1 v minor change and add a newline [here](https://github.com/alphagov/govuk-chat/pull/388/commits/c5eca88a7db812d52e6a30f094cd950bd246b6e4#diff-c505b93426c982f194635ea1dbfe541a79c8d17153485f6a02630182526f4dd0R114). Just because it's such a minor change it's quicker than leaving a comment (sorry i missed it in the previous round of comments).

But yea i would approve but i opened the PR so i can't. So if you're happy with it then please do approve and i'll merge it

exfalsoquodlibet (approved):
Thanks David! 🚀

PR #389, opened 2025-07-17
Add test to ensure question question routing labels match prompts config

davidgisbey:
We had an issue where we bumped the private gem with some breaking changes, expecting CI to catch it, so we could apply changes there.

Instead it went green and got merged with the breaking changes. The reason it didn't blow up is because we weren't using the removed question routing labels in our pipeline tests. So we don't try and dig for the prompt in the config anywhere in our tests.

We have a test that checks the question routing labels enum values and the question_routing_labels.yml config are in sync. I've added a similar test to check that our prompt config is also in sync.

### DIFF
--- a/spec/models/answer_spec.rb
+++ b/spec/models/answer_spec.rb
@@ -211,6 +211,19 @@ RSpec.describe Answer do
     expect(label_config.keys).to match_array(enum_values)
   end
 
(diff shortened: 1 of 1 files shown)


jackbot (approved):
Nice one :+1:

PR #392, opened 2025-07-21
Import tag component

AshGDS:
- Import the tag component
- Was missed in #384
- Will fix this tag having no styles 
image

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -16,6 +16,7 @@
 @import "govuk_publishing_components/components/radio";
 @import "govuk_publishing_components/components/skip-link";
 @import "govuk_publishing_components/components/success-alert";
+@import "govuk_publishing_components/components/tag";
 @import "govuk_publishing_components/components/warning-text";
 
 @import "objects/conversation-layout";


JamesCGDS (approved):
Good spot, ty

kevindew:
Thanks for opening this ⭐ 

@JamesCGDS I don't think this has fixed it though as the stylesheet this was applied to is our application one whereas the screenshot is from the admin area. 

It seems surprising that we'd resolve this issue by adding the component in to our base ones though since we don't use it directly, I'd have expected [_layout_header.scss](https://github.com/alphagov/govuk_publishing_components/blob/b63c2b6605912f5b4b2f9ee4ce0057deacfd346e/app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss#L1-L3) to include it as per other component dependencies in publishing_components? Could we resolve it there @AshGDS or was that rejected for a reason?

kevindew:
> @kevindew Ah my mistake, it should be imported into `admin.scss` instead - who wants to do the PR for that? (Happy to if asked)
> 
> I think you would need to configure your application to support individual component stylesheets in order to remove the `@import` statements in the repo: https://github.com/alphagov/govuk_publishing_components/blob/main/docs/set-up-individual-component-css-loading.md

Thanks, but do what dig a bit more as I'm a bit confused.

Maybe I'm missing something but it looks to me like there is now a bug with the `_layout_header.scss` component as the only way that can work is if you separately import the tag component - which is inconsistent with the approach of the layout_header component and other components, where dependent components are imported. So feels to me that amending our stylesheets is just a plaster over the broader bug (given tag isn't a component in [component_support](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/stylesheets/govuk_publishing_components/component_support.scss) or [govuk_frontend_support](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/stylesheets/govuk_publishing_components/govuk_frontend_support.scss)). So rather than doing this should we just amend _layout_header.scss to import the tag component?

With individual component stylesheets there would still be a need to import it as it's not on `_layout_header.scss`?

AshGDS:
Thanks @kevindew - my understanding of the assets pipeline isn't the best so I might be wrong!

Is `layout_header` maybe an outlier here? No other components seem to have an `@import "govuk_publishing_components/components` statement in their `.scss` - for example [this component](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_contents_list_with_body.html.erb) calls two components but [doesn't import them anywhere in the SASS](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/stylesheets/govuk_publishing_components/components/_contents-list-with-body.scss) , and that's because the [child components import their own SASS themselves within the template](https://github.com/alphagov/govuk_publishing_components/blob/8b1be49d16d8dfaa113916e29a3915e1be85bc11/app/views/govuk_publishing_components/components/_contents_list.html.erb#L2)

My understanding was that if the gem's AssetHelper was used in your rendering app, then those `add_gem_component_stylesheet` statements would handle the imports for you: https://github.com/alphagov/govuk_publishing_components/blob/b63c2b6605912f5b4b2f9ee4ce0057deacfd346e/lib/govuk_publishing_components/app_helpers/asset_helper.rb#L39-L59 / https://github.com/alphagov/frontend/pull/3342

I can ask one of the devs involved in the individual CSS work if needed as I can see what you mean by the import being missing in `_layout_header.scss`, but I'm not sure it should be needed

AshGDS:
Thanks @kevindew - yep I can fix those `_layout-header` imports in the gem 👍

PR #394, opened 2025-07-21
Move tag component stylesheet to admin area

kevindew:
This was mistakenly added to the application.scss file when it is expected to only be needed in the admin area where the layout-header component is used.

It is needed because recent changes to the layout-header component have removed the import of a tag stylesheet which now needs to be imported manually [1].

Before:
Screenshot 2025-07-21 at 19 27 21

After:
Screenshot 2025-07-21 at 19 26 53

[1]: https://github.com/alphagov/govuk_publishing_components/pull/4914


AshGDS (approved):
LGTM 👍

PR #403, opened 2025-07-24
Add topic classification via post answer composition workflow

davidgisbey:
## Description

This spikes out implementing topic tagging for questions. This is run post answer composition because we need to run tagging on the rephrased question if the question was rephrased.

It currently points towards this [PR](https://github.com/alphagov/govuk_chat_private/pull/30) on the private repo which adds the required config for the LLM call. It should be noted that this is based on some info provided by David Gillespie and isn't the finalised version. I created the system prompt just to get something working, but the topics and tool call were largely provided by David. Def still subject to change though.

This PR:
- adds an AnswerAnalysis model and table to the db which will capture various metadata we want for analysis. For this it captures
  - primary topic
  - secondary topic
  - metics
  - llm respones
- adds an AnswerInsightsJob which is called at the end of answer composition
- adds the AnswerInsight::TopicTagger which  is called within the above job. This makes the LLM call to tag topics to the question and persists an AnswerAnalysis on the recently composed anser
- adds the topics, metrics and llm responses to the admin question show page
- adds filter to the questions index page to filter on primary and secondary topic

It should be noted that we only want to tag topics for answers which successfully completed the pipeline and weren't triggered by malicious user behaviour (jailbreak guardrails). So these have been filtered out 

### Question index

image

image

### Question show 

image

image

image

### DIFF
--- a/app/helpers/admin/questions_helper.rb
+++ b/app/helpers/admin/questions_helper.rb
@@ -201,6 +201,19 @@ module Admin
         ]
       end
 
+      if answer&.analysis&.primary_topic.present?
+        rows << [
+          {
+            field: "Primary topic",
+            value: answer.analysis.primary_topic.humanize,
+          },
+          {
+            field: "Secondary topic",
+            value: answer.analysis.secondary_topic&.humanize,
+          },
+        ]
+      end
+
       rows.flatten
     end
 
--- a/app/jobs/answer_topics_job.rb
+++ b/app/jobs/answer_topics_job.rb
@@ -1,0 +1,14 @@
+class AnswerTopicsJob < ApplicationJob
+  MAX_RETRIES = 5
+  queue_as :default
+  retry_on Anthropic::Errors::APIError, wait: 1.minute, attempts: MAX_RETRIES
+
+  def perform(answer_id)
+    answer = Answer.includes(:analysis, question: :conversation).find_by(id: answer_id)
+
+    return logger.warn("No answer found for #{answer_id}") unless answer
+    return logger.warn("Answer #{answer_id} has already been tagged with topics") if answer.analysis&.primary_topic.present?
+
+    AnswerAnalysisGeneration::TopicTagger.call(answer)
+  end
+end
--- a/app/jobs/answer_topics_job.rb
+++ b/app/jobs/answer_topics_job.rb
@@ -8,6 +8,9 @@ class AnswerTopicsJob < ApplicationJob
 
     return logger.warn("No answer found for #{answer_id}") unless answer
     return logger.warn("Answer #{answer_id} has already been tagged with topics") if answer.analysis&.primary_topic.present?
+    unless answer.eligible_for_topic_analysis?
+      return logger.info("Answer #{answer_id} is not eligible for topic analysis")
+    end
 
     AnswerAnalysisGeneration::TopicTagger.call(answer)
   end
--- a/app/jobs/compose_answer_job.rb
+++ b/app/jobs/compose_answer_job.rb
@@ -13,5 +13,7 @@ class ComposeAnswerJob < ApplicationJob
     rescue ActiveRecord::RecordNotUnique
       logger.warn("Already an answer created for #{question_id}")
     end
+
+    AnswerTopicsJob.perform_later(answer.id) if answer.persisted?
   end
 end
--- a/app/models/answer_analysis.rb
+++ b/app/models/answer_analysis.rb
@@ -1,3 +1,5 @@
 class AnswerAnalysis < ApplicationRecord
+  include LlmCallsRecordable
+
   belongs_to :answer
 end
--- a/app/models/answer_analysis.rb
+++ b/app/models/answer_analysis.rb
@@ -1,0 +1,3 @@
+class AnswerAnalysis < ApplicationRecord
+  belongs_to :answer
+end
--- a/app/models/answer_analysis.rb
+++ b/app/models/answer_analysis.rb
@@ -2,4 +2,12 @@ class AnswerAnalysis < ApplicationRecord
   include LlmCallsRecordable
 
   belongs_to :answer
+
+  scope :exportable, lambda { |start_date, end_date|
+    where(created_at: start_date...end_date)
+  }
+
+  def serialize_for_export
+    as_json(except: :llm_responses).merge("llm_responses" => llm_responses.to_json)
+  end
 end
--- a/app/views/admin/questions/_question_filters.html.erb
+++ b/app/views/admin/questions/_question_filters.html.erb
@@ -179,6 +179,36 @@
     end,
   } %>
 
+  <%= render "govuk_publishing_components/components/select", {
+    id: "primary_topic",
+    name: "primary_topic",
+    label: "Primary topic",
+    heading_size: "s",
+    full_width: true,
+    options: [{text: "", value: ""}] + Rails.configuration.question_topics.map do |topic|
+      {
+       text: topic.humanize,
+       value: topic,
+       selected: params[:primary_topic] == topic,
+      }
+    end,
+  } %>
+
+  <%= render "govuk_publishing_components/components/select", {
+    id: "secondary_topic",
+    name: "secondary_topic",
+    label: "Secondary topic",
+    heading_size: "s",
+    full_width: true,
+    options: [{text: "", value: ""}] + Rails.configuration.question_topics.map do |topic|
+      {
+       text: topic.humanize,
+       value: topic,
+       selected: params[:secondary_topic] == topic,
+      }
+    end,
+  } %>
+
   <%= render "govuk_publishing_components/components/button", {
     text: "Filter",
     margin_bottom: true,
--- a/config/application.rb
+++ b/config/application.rb
@@ -66,7 +66,6 @@ module GovukChat
     config.answer_statuses = Hashie::Mash.new(YAML.load_file("#{__dir__}/answer_statuses.yml"))
     config.question_routing_labels = Hashie::Mash.new(YAML.load_file("#{__dir__}/question_routing_labels.yml"))
     config.search = Hashie::Mash.new(YAML.load_file("#{__dir__}/search.yml"))
-
     config.action_dispatch.rescue_responses.merge!(
       "Search::ChunkedContentRepository::NotFound" => :not_found,
       "ThrottledRequest" => :too_many_requests,
@@ -85,5 +84,11 @@ module GovukChat
     config.bigquery_dataset_id = ENV["BIGQUERY_DATASET"]
 
     config.answer_strategy = ENV.fetch("ANSWER_STRATEGY", "claude_structured_answer")
+
+    config.question_topics = GovukChatPrivate.config
+                                             .llm_prompts.claude
+                                             .topic_tagger
+                                             .dig("tool_spec", "input_schema", "$defs", "govuk_topic_tags", "enum")
+                                             .sort
   end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.0].define(version: 2025_07_14_145727) do
+ActiveRecord::Schema[8.0].define(version: 2025_07_30_115215) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
   enable_extension "pg_catalog.plpgsql"
@@ -22,6 +22,17 @@ ActiveRecord::Schema[8.0].define(version: 2025_07_14_145727) do
   create_enum "conversation_source", ["web", "api"]
   create_enum "guardrails_status", ["pass", "fail", "error"]
   create_enum "question_routing_label", ["about_mps", "advice_opinions_predictions", "character_fun", "genuine_rag", "gov_transparency", "greetings", "harmful_vulgar_controversy", "multi_questions", "negative_acknowledgement", "non_english", "personal_info", "positive_acknowledgement", "vague_acronym_grammar", "unclear_intent", "requires_account_data"]
+
+  create_table "answer_analyses", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.string "primary_topic"
+    t.string "secondary_topic"
+    t.jsonb "metrics"
+    t.jsonb "llm_responses"
+    t.uuid "answer_id", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["answer_id"], name: "index_answer_analyses_on_answer_id", unique: true
+  end
 
   create_table "answer_feedback", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
     t.uuid "answer_id", null: false
@@ -139,6 +150,7 @@ ActiveRecord::Schema[8.0].define(version: 2025_07_14_145727) do
     t.datetime "updated_at", null: false
   end
 
+  add_foreign_key "answer_analyses", "answers", on_delete: :cascade
   add_foreign_key "answer_feedback", "answers", on_delete: :cascade
   add_foreign_key "answer_sources", "answers", on_delete: :cascade
   add_foreign_key "answers", "questions", on_delete: :cascade
--- a/lib/answer_analysis_generation/topic_tagger.rb
+++ b/lib/answer_analysis_generation/topic_tagger.rb
@@ -12,6 +12,7 @@ class AnswerAnalysisGeneration::TopicTagger
   def call
     start_time = Clock.monotonic_time
 
+    raise "Answer #{answer.id} is not eligible for topic analysis" unless answer.eligible_for_topic_analysis?
     raise "Topics already generated for answer #{answer.id}" if answer.analysis&.primary_topic.present?
 
     response = anthropic_bedrock_client.messages.create(
--- a/lib/answer_analysis_generation/topic_tagger.rb
+++ b/lib/answer_analysis_generation/topic_tagger.rb
@@ -1,0 +1,82 @@
+class AnswerAnalysisGeneration::TopicTagger
+  delegate :logger, to: Rails
+
+  attr_reader :answer
+
+  def self.call(...) = new(...).call
+
+  def initialize(answer)
+    @answer = answer
+  end
+
+  def call
+    start_time = Clock.monotonic_time
+
+    raise "Topics already generated for answer #{answer.id}" if answer.analysis&.primary_topic.present?
+
+    response = anthropic_bedrock_client.messages.create(
+      messages:,
+      system: [
+        { type: "text", text: system_prompt, cache_control: { type: "ephemeral" } },
+      ],
+      model: BedrockModels.model_id(:claude_sonnet),
+      tools:,
+      tool_choice: { type: "tool", name: tools.first[:name] },
+      **inference_config,
+    )
+
+    analysis = answer.build_analysis
+    analysis.primary_topic = response[:content][0][:input][:primary_topic]
+    analysis.secondary_topic = response[:content][0][:input][:secondary_topic]
+    analysis.assign_metrics("topic_tagger", build_metrics(response, start_time))
+    analysis.assign_llm_response("topic_tagger", response.to_h)
+
+    analysis.save!
+  end
+
+private
+
+  def anthropic_bedrock_client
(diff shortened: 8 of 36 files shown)


config/answer_topic.yml
```diff
@@ -0,0 +1,154 @@
+system_prompt: |
```
davidgisbey, line 1: obviously this file will live in the private repo, but i've just added it here for now so it's easy to see/review in one place.

app/jobs/compose_answer_job.rb
```diff
@@ -10,6 +10,7 @@ def perform(question_id)
 
     begin
       answer.save!
+      AnswerTopic::Tagger.call(Answer.includes(%i[topic question]).find(answer.id))
```
kevindew, line 13: I think we probably want this outside the begin/rescue since the rescue isn't really for this (and would be very confusing if it triggered).

I imagine we also want this to run against a question that has timed out? it could even be that we want to change the way guardclauses are done so it doesn't exit early on question.answer but instead checks for question.answer, if none generates an answer, then answer.topic, if non tag a topic.

lib/answer_topic/tagger.rb
```diff
@@ -0,0 +1,14 @@
+class AnswerTopic::Tagger
```
kevindew, line 1: So I think we should avoid using AnswerTopic as the root namespace here since it's already a model and thus can't be a module.

Perhaps AnswerAnalysis would be a good route namespace and this could be TopicTagger ?

lib/answer_topic/tagger.rb
```diff
@@ -0,0 +1,14 @@
...
+  def self.call(answer)
+    answer_strategy = answer.question.answer_strategy.to_sym
```
kevindew, line 3: As discussed verbally, it'd be good if this didn't need to be aware of answer strategy - we'll just use claude as the tagger until we need to consider otherwise.

I imagine this may have a knock on that we might need to set-up heroku OpenAI env to have AWS creds if it doesn't already.

lib/answer_topic/claude/tagger.rb
```diff
@@ -0,0 +1,90 @@
...
+    .join("\n")
+  end
+
+  def topics_config
+    Rails.application.config.answer_topic
+  end
+
+  def inference_config
+    {
+      max_tokens: 4096,
```
kevindew, line 82: Not something we'd need to change right now, but we will want this to likely be quite a lot smaller for this task

kevindew (commented):
Looks really good, just a few implementation comments. Happy to discuss more.

As an aside, I do wonder if we might be reproducing the same sort of Claude call a model, log metrics and llm_response a few times. Not sure if there's a potentially helper class or function potentially looming.

app/jobs/compose_answer_job.rb
davidgisbey: yea the reason i haven't implemented the error timeout one is i'm not sure we would want to run it on a canned answer? So i've held off doing anymore until i get a clear answer which answer statuses we want to add topics too. But i'll def keep it in mind.

Good shout about moving it out. I did end have 

```
AnswerTopic::Tagger.call(Answer.includes(%i[topic question]).find(answer.id)) if answer.persisted?
```

below but moved it up. Should've left it

lib/answer_topic/tagger.rb
davidgisbey: good shout will do

lib/answer_topic/tagger.rb
davidgisbey: Good shout. I'll make sure i change this post chat with David. (or today if i have time after the BigQuery stuff.)

lib/answer_topic/claude/tagger.rb
davidgisbey: Yea agreed. I just added this as a placeholder until David feeds back on his work

app/jobs/answer_insights_job.rb
```diff
@@ -0,0 +1,10 @@
+class AnswerInsightsJob < ApplicationJob
```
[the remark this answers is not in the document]
davidgisbey: Yea sounds good. I added this prior to knowing whether we were going with an async approach or not, but given we're going to call multiple jobs concurrently i think we can change this now

lib/answer_insights/topic_tagger.rb
```diff
@@ -0,0 +1,96 @@
+class AnswerInsights::TopicTagger
```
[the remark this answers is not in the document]
davidgisbey: Yea i was trying to find a name that covers generating analysis and manual eval work, but it might be that they have sep models. 

I have a preference for `AnswerAnalysisGeneration` out of the two. `AnswerAnalysing` feels a bit clunkier

app/views/admin/questions/_question_filters.html.erb
```diff
@@ -179,6 +179,36 @@
     end,
   } %>
 
+  <%= render "govuk_publishing_components/components/select", {
+    id: "primary_topic",
+    name: "primary_topic",
+    label: "Primary topic",
+    heading_size: "s",
+    full_width: true,
+    options: [{text: "", value: ""}] + Rails.configuration.govuk_chat_private.llm_prompts.claude.topic_tagger["topics"].keys.map do |topic|
```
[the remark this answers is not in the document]
davidgisbey: Yea it's an interesting one. We don't use answer_statuses at all in the prompts. All the logic for it is in the public codebase so that's why i leaned towards the private repo. Having said that i don't think it's something we particularly case about being public.

I guess what we need to consider is. Do we want all info used in the prompt to be private? In which case it should live in `govuk_chat_private.analysis_topics`. If not then we can just add a config file in this repo.

I'm leaning towards the private repo. Just becuase it feels like it's easier to reason what should live where when  anything directly used in prompts lives there. 

Otherwise i feel like each time we add something new like this we will need to discuss where it lives. Wdyt?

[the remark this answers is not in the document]
kevindew: Yeah a better example I should have picked is question_routing_labels.yml as something where we have bits in both places - so probably a better analogue.

Yeah we could go all for the private repo until we see evidence of a problem, at least that means changes could be atomic.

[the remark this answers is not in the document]
davidgisbey: ah right i see what you mean. Maybe it should just live in the public one then if there's prior art of it being done. It doesn't really feel like it needs to live in the private repo and we do use it several times in the admin UI.

And i can't imagine any added security concerns considering we're not running it on Jailbreaks.

davidgisbey:
@kevindew i ended up adding the topics into config in this repo. I've left a couple of new things in the last few commits:

1. i've locked the answer before creating/updating the answer analysis in the 3rd to last commit
2. i've moved metric and llm_response assignment to a shared concern
3. i've added retry logic to the topics job

spec/models/answer_analysis_spec.rb
```diff
@@ -41,4 +41,37 @@
       expect(answer_analysis.serialize_for_export).to eq(expected_response)
     end
   end
+
+  describe "#assign_metrics" do
```
davidgisbey, line 45: these can probs live in a shared example, but for now i just did this until i got some feedback on what people thought about the abstraction.

lib/answer_analysis_generation/topic_tagger.rb
```diff
@@ -0,0 +1,100 @@
...
+        { type: "text", text: system_prompt, cache_control: { type: "ephemeral" } },
+      ],
+      model: BedrockModels::CLAUDE_SONNET,
+      tools:,
+      tool_choice: { type: "tool", name: tools.first[:name] },
+      **inference_config,
+    )
+
+    answer.reload.with_lock do
+      raise "Topics already generated for answer #{answer.id}" if answer.analysis&.primary_topic.present?
```
kevindew, line 30: Is this exception likely at all? I'm thinking it's probably super rare, i.e. could only be caused by somehow sidekiq getting duplicate jobs and running them concurrently, so probably fine. But did wonder if there was a likely scenario for it we'd not want it to polute our error logs and would potentially want it as something caught by the job.

kevindew (commented):
I've run out of time so only a quick review but think comments should be interesting at least

lib/answer_analysis_generation/topic_tagger.rb
```diff
@@ -0,0 +1,100 @@
...
+      system: [
+        { type: "text", text: system_prompt, cache_control: { type: "ephemeral" } },
+      ],
+      model: BedrockModels::CLAUDE_SONNET,
+      tools:,
+      tool_choice: { type: "tool", name: tools.first[:name] },
+      **inference_config,
+    )
+
+    answer.reload.with_lock do
```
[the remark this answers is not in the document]
davidgisbey: So the reason i locked the answer here which might be premature is that soon we're going to be running multiple jobs with llm calls concurrently

1. topics
2. question intent
3. question category (general, specific etc.)

Each of these could finish before the other 2 and would then need to either

a. create the analysis
b. update the existing analysis

so i figured locking the answer would be the simplest solution. I guess we could do 

```
 analysis = answer.analysis || answer.build_analysis
```

and catch all errors for retries so if `raise "Analysis already created for answer #{answer.id}"` is raised then it'll just get retried.

lib/answer_analysis_generation/topic_tagger.rb
davidgisbey: Yea it really shouldn't occur. Def happy to take it out.

kevindew: Fine (and probably better) to keep in if it's a last ditch state check. 

Just wanted to check that we should have something that is less catasrophic if the scenario is actually likely.

lib/answer_analysis_generation/topic_tagger.rb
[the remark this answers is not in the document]
kevindew: So I expect we'd do something different if we have multiple jobs. What I expect that would look like is:

```
# find_or_create_by! handles a RecordNotUnique exception
analysis = AnswerAnalysis.find_or_create_by!(answer_id: answer.id)

# do the analysis

analysis.with_lock do
  # any guard clauses that this analysis already exists
  analysis.update!(# however it has changed)
end
```

app/jobs/answer_topics_job.rb
```diff
@@ -1,5 +1,7 @@
 class AnswerTopicsJob < ApplicationJob
+  MAX_RETRIES = 5
   queue_as :default
+  retry_on Anthropic::Errors::APIError, wait: 1.minute, attempts: MAX_RETRIES
```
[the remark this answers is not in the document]
davidgisbey: Ah yea good point. 
``` pry(main)> AnswerTopicsJob.sidekiq_options
=> {"retry" => true, "queue" => "default"}
```
so we can either toggle them off in the job via `sidekiq_options retry: false` or we could just retry via Sidekiq. Do you have a preference?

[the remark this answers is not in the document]
kevindew: I think turning them off at sidekiq is better to avoid two systems. The fact there are two systems is a bit of a source of pain (the ActiveJob retries won't end up in the Sidekiq retry queue but in the scheduled queue)

lib/answer_analysis_generation/topic_tagger.rb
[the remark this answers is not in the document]
davidgisbey: Okay cool. I went with your first suggestion for now since we can move to the second when the need arises.

db/schema.rb
```diff
@@ -23,6 +23,17 @@
...
 
+  create_table "answer_analyses", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.string "primary_topic"
+    t.string "secondary_topic"
+    t.jsonb "metrics"
+    t.jsonb "llm_responses"
+    t.uuid "answer_id", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["answer_id"], name: "index_answer_analyses_on_answer_id", unique: true
```
exfalsoquodlibet, line 34: Maybe I am missing it as my understanding of Ruby is limited, but could we also record the question that was actually tagged?

davidgisbey: So it's associated with the answer, since we pass the answer into the tagger once it's persisted in the database so we have access to the rephrased question (which lives on the answer).

But the answer is associated with the question. So we can just do `analyisis.answer.question` to get access to the question that was tagged.

kevindew:
Maybe we can grab some time (after standup perhaps) to talk about naming?

db/schema.rb
exfalsoquodlibet: Thanks, David, I was tied up in back-to-back meetings.

I had a quick look at the DB schema and asked Claude to represent it in a diagram. I can see that rich metadata is stored using JSONB for `metrics` and `llm_response`, which makes sense since these fields need to be flexible enough to handle evolving data structures.

Just to confirm my understanding:

* **`metrics`** captures something like:

  ```ruby
  def build_metrics(response, start_time)
    {
      duration: Clock.monotonic_time - start_time,
      llm_prompt_tokens: BedrockModels.claude_total_prompt_tokens(response[:usage]),
      llm_completion_tokens: response[:usage][:output_tokens],
      llm_cached_tokens: response[:usage][:cache_read_input_tokens],
      model: response[:model],
    }
  end
  ```

* **`llm_response`** stores the raw JSON response from the LLM call (e.g. from the TopicTagger), following a structure like:

  ```ruby
  expected_llm_response = {
    "id" => "msg-id",
    "content" => [
      {
        "id" => "tool-use-id",
        "input" => {
          "primary_topic" => "business",
          "secondary_topic" => "benefits",
          "confidence" => "high",
          "reasoning" => "reason"
        },
        "name" => "tagger_reasoning",
        "type" => "tool_use"
      }
    ],
    "model" => BedrockModels::CLAUDE_SONNET,
    "role" => "assistant",
    "stop_reason" => "tool_use",
    "type" => "message",
    "usage" => {
      "cache_read_input_tokens" => 20,
      "input_tokens" => 10,
      "output_tokens" => 20
    },
  }
  ```

This is just me trying to understand exactly what we’re storing and how, nothing that should block this PR from merging.

app/helpers/admin/questions_helper.rb
```diff
@@ -201,6 +201,20 @@ def question_show_summary_list_rows(question, answer, question_number, total_que
         ]
       end
 
+      if answer&.analysis&.primary_topic.present?
+        rows << {
+          field: "Primary topic",
+          value: answer.analysis.primary_topic.humanize,
+        }
+
+        if answer.analysis.secondary_topic.present?
```
kevindew, line 210: I think we should probably show this as empty if we have a primary topic. 

On the basis that this not actually having a result is still an outcome of analysis.

app/jobs/answer_topics_job.rb
```diff
@@ -0,0 +1,17 @@
+class AnswerTopicsJob < ApplicationJob
+  MAX_RETRIES = 5
+  queue_as :default
```
kevindew, line 3: I think we might have some lingering work to do on prioritising job execution - so this can't block answer composition. We can do this post merge though

app/models/answer_analysis.rb
```diff
@@ -0,0 +1,13 @@
...
+  include LlmCallsRecordable
+
+  belongs_to :answer
+
+  scope :exportable, lambda { |start_date, end_date|
+    where(created_at: start_date...end_date)
+  }
+
+  def serialize_for_export
+    as_json.merge("llm_responses" => llm_responses.to_json)
```
kevindew, line 11: A micro-optimisation is to exclude llm_responses from as_json so you don't have the penalty of that being serialised for no reason.

config/application.rb
```diff
@@ -66,6 +66,7 @@ class Application < Rails::Application
     config.answer_statuses = Hashie::Mash.new(YAML.load_file("#{__dir__}/answer_statuses.yml"))
     config.question_routing_labels = Hashie::Mash.new(YAML.load_file("#{__dir__}/question_routing_labels.yml"))
     config.search = Hashie::Mash.new(YAML.load_file("#{__dir__}/search.yml"))
+    config.topics = YAML.load_file("#{__dir__}/topics.yml")
```
kevindew, line 69: Could we call this question_topics - just to get a prefix in there early as a domain concept to avoid clashes with any other type of topic. 

It doesn't necessary to change this on the Analysis model as we already have the context of it being on answer

kevindew: Also, I know I'm treading old ground but just wondering given we're using the list from the prompt at the moment whether we should find a way to avoid having two lists to maintain for now (given there's not any deviation) 

I think perhaps the way that seems least painful is if we populated this config variable with data from the private gem (I think dependencies are available here)? Could even have a test that asserts it's a non empty array. That strikes me as maybe a way to avoid duplication but not have deeply mixed domain concepts.

davidgisbey: Okay great. I've used the topics in input_schema of the private repo and just added a basic test to check that there's an array containing something when you pulll the schema configs enum.

app/jobs/answer_topics_job.rb
davidgisbey: Sounds good. I'll card something up for it once this has been merged.

app/jobs/answer_topics_job.rb
```diff
@@ -0,0 +1,17 @@
...
+  def perform(answer_id)
+    answer = Answer.includes(:analysis, question: :conversation).find_by(id: answer_id)
+
+    return logger.warn("No answer found for #{answer_id}") unless answer
+    return logger.warn("Answer #{answer_id} has already been tagged with topics") if answer.analysis&.primary_topic.present?
+    unless answer.eligible_for_topic_analysis?
+      return logger.info("Answer #{answer_id} is not eligible for topic analysis")
+    end
+
+    AnswerAnalysisGeneration::TopicTagger.call(answer)
```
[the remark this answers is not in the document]
davidgisbey: yup makes sense 👍 i guess down the line we just need to find_or_init and lock the instance.

davidgisbey:
@kevindew thanks for the review. Just getting to look at it now as i working on gradio stuff today, but i'll prioritise getting this done next

app/views/admin/questions/_question_filters.html.erb
```diff
@@ -1,6 +1,11 @@
 <%
   conversation = filter.conversation
   signon_user = filter.signon_user
+  question_topics = Rails.configuration
+                         .govuk_chat_private
+                         .llm_prompts
+                         .claude
+                         .topic_tagger.dig("tool_spec", "input_schema", "$defs", "govuk_topic_tags", "enum")
```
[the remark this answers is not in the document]
davidgisbey: So it doesn't have access to `config.govuk_chat_private` here, but i can do 

```
    config.question_topics = GovukChatPrivate.config
                                                                              .llm_prompts.claude
                                                                              .topic_tagger
                                                                             .dig("tool_spec", "input_schema", "$defs", "govuk_topic_tags", "enum")
                                                                              .sort

```

or do it in an intialiser. Do you have a preference?

davidgisbey:
@kevindew  i've added some stuff to a TODO list

spec/models/answer_analysis_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  describe ".exportable" do
+    let(:new_answer_analysis) { create(:answer_analysis, created_at: 2.days.ago) }
+    let(:old_answer_analysis) { create(:answer_analysis, created_at: 4.days.ago) }
+
+    before do
+      new_answer_analysis
+      old_answer_analysis
+    end
```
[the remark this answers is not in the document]
kevindew: Ah sure - that warning makes sense. I'd say the most pragmatic approach for immediate way is to do:

```
let!(:exported_answer_analysis) { create(:answer_analysis, created_at: 2.days.ago) }
before { create(:answer_analysis, created_at: 4.days.ago) }
```

I imagine if we refactor things more for a shared we may want to explicitly assert the item that isn't there. 

e.g. 

```
expect(something).to include(exportable_answer_analysis)
expect(something).not_to include(exportable_answer_analysis)
```

but doesn't feel right for this exact test

kevindew:
Great - think we're just about there on this now

spec/models/answer_analysis_spec.rb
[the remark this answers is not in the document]
davidgisbey: okay cool. have sorted that

davidgisbey:
Thanks @kevindew i've made that final change. I'm now working on the follow up tasks. I'll card them up now.

PR #411, opened 2025-07-28
Persist LLM response for unanswerable questions

davidgisbey:
We're not persisting the LLM response for unanswerable questions during structured answer composition. This change ensures that the LLM response is assigned prior to the check to see if the question is answerable since we always want to persist it.


kevindew (approved):
Nice catch and test

PR #419, opened 2025-07-30
Update homepage

JamesCGDS:
## What
This PR removes the messages animation from the homepage (agreed with designers) and adds the new speech bubble icon with an animating gradient. For users that have activated the `prefers-reduced-motion` media query, the animating icon will be hidden and a static version will be shown.

Tested in latest versions of Chrome, Firefox, Safari, and Edge and everything looks like it's working okay. In browsers that don't support svg animations (e.g. Edge 16-18), the svg is just static and the gradient doesn't animate.

## Why
[Trello card](https://trello.com/c/LOsbizjI/2655-implement-css-gradient-animations)

### Desktop

| Homepage - before | Homepage - after |
|------------------------------|----------------------------------------|
| image | image |

### Mobile

| Homepage - before | Homepage - after |
|------------------------------|----------------------------------------|
| image | image |


app/assets/images/components/chat_introduction_title/govuk-chat-introduction-title-icon.svg
kevindew, line 1: Ooh that's clever that the SVG itself animates (and without any inline styles)

kevindew (commented):
Looks good to me, I did wonder if perhaps it was a bummer that we serve both images to everyone (with the inline aspect) but given collectively they're less than 5KB it doesn't seem worth a concern.

PR #423, opened 2025-07-31
Implement text area

JamesCGDS:
## What
This PR switches out the input element for an expanding textarea element. The implementation was based on this [blog post](https://chriscoyier.net/2023/09/29/css-solves-auto-expanding-textareas-probably-eventually/).

I'm conscious that the placement of the `Send` button might be preferred at the bottom/aligned with the bottom of the text area. This seems to be a little tricky to achieve due to how our DOM is structured and it could involve a rework of the component. Therefore, I've opted for omitting that work from this PR to avoid holding it up. Happy to discuss.

## Why
[Trello card](https://trello.com/c/DqBmf2u3/2656-implement-text-area)

### Desktop

| Chat UI input - before | Chat UI input - after |
|------------------------------|----------------------------------------|
| image | image |

### Mobile

| Chat UI input - before | Chat UI input - after |
|------------------------------|----------------------------------------|
| image | image |

### DIFF
--- a/app/assets/javascripts/components/question-form.js
+++ b/app/assets/javascripts/components/question-form.js
@@ -10,7 +10,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     constructor (module) {
       this.module = module
       this.form = module.querySelector('.js-question-form')
-      this.input = module.querySelector('.js-question-form-input')
+      this.textarea = module.querySelector('.js-question-form-textarea')
+      this.textareaWrapper = module.querySelector('.js-question-form-textarea-wrapper')
       this.button = module.querySelector('.js-question-form-button')
       this.buttonResponseStatus = module.querySelector('.js-question-form-button__response-status')
       this.errorsWrapper = module.querySelector('.js-question-form-errors-wrapper')
@@ -30,6 +31,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       this.module.dispatchEvent(new Event('init'))
 
       new window.GOVUKFrontend.CharacterCount(this.module) // eslint-disable-line no-new
+
+      this.enableTextareaResizing(this.textareaWrapper)
+      this.submitTextAreaOnEnter(this.textarea)
     }
 
     handleSubmit (event) {
@@ -41,14 +45,14 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
       const errors = []
 
-      if (this.input.value.trim() === '') {
+      if (this.textarea.value.trim() === '') {
         errors.push(this.module.dataset.presenceErrorMessage)
       }
 
       this.replaceErrors(errors)
 
       const maxlength = parseInt(this.module.dataset.maxlength, 10)
-      const exceedsMaxLength = this.input.value.length > maxlength
+      const exceedsMaxLength = this.textarea.value.length > maxlength
 
       if (errors.length || exceedsMaxLength) {
         event.preventDefault()
@@ -63,7 +67,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
     handleQuestionAccepted () {
       this.disableControls()
-      this.resetInput()
+      this.resetTextarea()
       this.handleButtonResponseStatus(this.buttonResponseStatus.dataset.loadingAnswerText)
     }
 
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/components/question-form.js
```diff
@@ -129,31 +133,46 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+
+    enableTextareaResizing (textareaWrapper) {
+      this.textarea.addEventListener('input', () => {
+        textareaWrapper.dataset.replicatedValue = this.textarea.value
+      })
+    }
+
+    submitTextAreaOnEnter (textarea) {
+      textarea.addEventListener('keypress', e => {
+        if (e.key === 'Enter') {
```
kevindew, line 170: I think we probably also want some handling for shift + enter and command + enter to actually get a new line. I'm not sure what/where the agreed convention is (I imagine we may have to do some research) but it's super common that they work (i.e. try chatgpt.com or claude.ai).

I imagine a follow up task (which we should probably do regardless) is some admin area work to work out what happens if we get a question with new lines in it and whether it renders in the admin area ok.

kevindew: I got a couple of things out of ChatGPT about it: https://chatgpt.com/s/t_688bb32697508191892cfd49aced2fb8 and https://chatgpt.com/s/t_688bb34504a881919a231e8916bc2850

app/assets/stylesheets/components/_question-form.scss
```diff
@@ -68,6 +71,22 @@
...
 
+.app-c-question-form__textarea-character-count-wrapper {
+  display: grid;
+
+  &::after {
+    content: attr(data-replicated-value) ""; // Note the weird space! Needed to preventy jumpy behavior
+    white-space: pre-wrap;  // This is how textarea text behaves
+    visibility: hidden; // Hidden from view, clicks, and screen readers
+    grid-area: 1 / 1 / 2 / 2;
+    max-height: 250px; // Set a max-height to prevent the textarea from growing too much
```
kevindew, line 82: Should we consider viewport height as part of this? Also 250px does feel potentially a bit on the small side for some bigger screens.

app/assets/stylesheets/components/_question-form.scss
```diff
@@ -68,6 +71,22 @@
   }
 }
 
+.app-c-question-form__textarea-character-count-wrapper {
+  display: grid;
+
+  &::after {
```
kevindew, line 77: It looks the height of this is a bit off as I end up with a scroll bar and would expect the growing to not show a scroll bar (again as per ChatGPT or Claude - although neither use a textarea so might not be the perfect example implementation wise but correct ux)

Jul-31-2025 19-24-55

kevindew: Oh I notice the one you linked to in your description doesn't have the scrolling issue

JamesCGDS:
Thanks for taking a look @kevindew, really appreciate the feedback (and apologies it's very much a half-baked solution atm). I'll be taking a look into tweaking/fixing things tomorrow 👍

Also re testing, I was thinking of putting in some tests for the enter key press behaviour - will continue looking into this tomorrow as well but would be good to have a quick chat about it next week

app/assets/javascripts/components/question-form.js
```diff
@@ -129,31 +134,47 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+        textareaWrapper.dataset.replicatedValue = this.textarea.value
+      })
+    }
+
+    handleTextareaKeypress (textarea) {
+      textarea.addEventListener('keydown', e => {
+        // Submit form on enter; add newline on enter + shift key
+        if (e.key === 'Enter' && !e.shiftKey) {
+          e.preventDefault()
+          window.GOVUK.triggerEvent(this.form, 'submit')
```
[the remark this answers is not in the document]
JamesCGDS: `this.form.requestSubmit()` does read nicer but unfortunately it's [not supported](https://caniuse.com/?search=requestSubmit) by our [grade C browsers](https://github.com/alphagov/govuk-frontend#browser-and-assistive-technology-support) (e.g. Chrome 62 or Edge 16-18) so I went with the above approach. Happy to chat about it though if we want to look into alternatives a bit more

JamesCGDS:
Thanks for taking a look at this @kevindew, have responded to the comments

Regarding the new line UI issue, I may have to look into that as a follow up task - it's proving a little tricky to get things displaying exactly how we'd like them. I'm sure I can figure something out but I'm not sure if I'll get there by the end of the day

app/assets/javascripts/components/question-form.js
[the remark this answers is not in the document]
kevindew: Cool, not to worry any chance you could add a comment to that effect since it seems like something that could be easily changed without realising you made a conscious decision against the standard route?

PR #432, opened 2025-08-04
Add API middleware request specs

kevindew:
It adds a distinct request spec for middleware which adds tests for aspects of middleware - some wasn't tested before at integration level and some (rate limits) was tested a lot at each individual routes.

More details in individual commits.


davidgisbey:
PR looks good to me. Makes sense to test exhaustively once and then just test it's rate limited for each endpoint

PR #437, opened 2025-08-05
Add common GOV.UK security actions

kevindew:
I understand we were previously missing CodeQL as we were a private
repo, and must have missed the memo about dependency-review.


davidgisbey (approved):
lgtm

PR #438, opened 2025-08-05
Send rate limit usage to Prometheus for API user

jackbot:
This extends the current rate limiting middleware to log rate limit
usage for an API user.

We want some way of knowing how close API users are to their request
quota, and doing this via Prometheus/Grafana seems like a good way to
go.

So in the middleware we inspect the rate limit data we've got from
Rack::Attack. If the rate limit data is related to the throttling for
the API user's read methods (and it's a read method) then we log the
percentage of read requests remaining for that user in Prometheus. Same
for write requests.

We could log the raw number, i.e. the number of requests remaining, but
this would be a bit meaningless when looking at the chart in Grafana
unless you know the actual rate limit figures. It feels like logging the
percentage remaining means we can create a chart that's easily
understandable at a glance.

Using a percentage as the metric also means that if we change the rate
limits, we just need to change one number in this app and we won't need
to make any changes to the chart.

Note that we're just logging the percentage remaining for the API user here,
not the end user. It feels like it'd be tricky to create a readable chart for
all the end users. We'll only have one API user to begin with (the mobile app),
so the chart will be simple.

### DIFF
--- a/lib/api/rate_limit/middleware.rb
+++ b/lib/api/rate_limit/middleware.rb
@@ -31,10 +31,26 @@ class Api::RateLimit::Middleware
       period = throttle_info[:period]
       epoch_time = throttle_info[:epoch_time]
       expires_in = (period - (epoch_time % period)).to_i
+      percentage_used = ((count.to_f / limit) * 100).round(2)
+      user_name = env.fetch("warden").user.name
 
       headers["#{header_name_prefix}-RateLimit-Limit"] = limit.to_s
       headers["#{header_name_prefix}-RateLimit-Remaining"] = [limit - count, 0].max.to_s
       headers["#{header_name_prefix}-RateLimit-Reset"] = "#{expires_in}s"
+
+      if header_name_prefix == "Govuk-Api-User-Read"
+        PrometheusMetrics.gauge(
+          "rate_limit_api_user_read_percentage_used",
+          percentage_used,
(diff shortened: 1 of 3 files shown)


lib/api/rate_limit/middleware.rb
```diff
@@ -30,13 +30,36 @@ def call(env)
...
 
       headers["#{header_name_prefix}-RateLimit-Limit"] = limit.to_s
       headers["#{header_name_prefix}-RateLimit-Remaining"] = [limit - count, 0].max.to_s
       headers["#{header_name_prefix}-RateLimit-Reset"] = "#{expires_in}s"
+
+      if header_name_prefix == "Govuk-Api-User-Read" && read_method?(env)
+        PrometheusMetrics.gauge(
+          "rate_limit_api_user_read_percentage_used",
+          percentage_used,
+          { signon_user_id: discriminator },
```
kevindew, line 45: I think we might be able to get a more humanly recognisable value for this discriminator as we should have the users login details via env["warden"].

So I think we could go with either `env.fetch("warden").user.name` or `env.fetch("warden").user.email` - the former seems like it'd be nicer for graphs but with a collision risk (that I think is acceptable)

kevindew (commented):
Nice job - a couple of suggestions from me but looks good.

PR #441, opened 2025-08-06
Update healthchecks

jackbot:
https://trello.com/c/adNt5wAW/2634

* Adds a Bedrock healthcheck
* Makes use of the existing `GovukHealthcheck::RailsCache` to test the cache
* Makes the OpenAI healthcheck conditional

### DIFF
--- a/lib/bedrock_models.rb
+++ b/lib/bedrock_models.rb
@@ -1,6 +1,20 @@
 module BedrockModels
-  CLAUDE_SONNET = ENV.fetch("CLAUDE_SONNET_MODEL_ID", "eu.anthropic.claude-sonnet-4-20250514-v1:0").freeze
-  TITAN_EMBED_V2 = "amazon.titan-embed-text-v2:0".freeze
+  MODEL_IDS = {
+    claude_sonnet: ENV.fetch("CLAUDE_SONNET_MODEL_ID", "eu.anthropic.claude-sonnet-4-20250514-v1:0").freeze,
+    titan_embed_v2: "amazon.titan-embed-text-v2:0",
+  }.freeze
+
+  def self.model_id(model_name)
+    MODEL_IDS.fetch(model_name) { raise "Unknown Bedrock model name: #{model_name}" }
+  end
+
+  def self.expected_foundation_models
+    # Strip the "eu." prefix from the model name we use if it exists. We
+    # sometimes use this prefix in our model names if we're using cross-region
+    # inference for that model. But the Bedrock API returns the model names without
+    # this region prefix.
+    MODEL_IDS.values.map { it.sub(/^eu\./, "") }
+  end
 
   def self.claude_total_prompt_tokens(usage)
     usage[:input_tokens].to_i + usage[:cache_read_input_tokens].to_i + usage[:cache_write_input_tokens].to_i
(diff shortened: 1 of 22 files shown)


kevindew (commented):
Thanks for that, I've added a few comments on bits that might be worth chatting about

lib/healthcheck/memcache.rb
```diff
@@ -0,0 +1,27 @@
+module Healthcheck
+  class Memcache
```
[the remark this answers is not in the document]
jackbot: Good spot, didn't see the existing Rails cache healthcheck. I've used that instead.

lib/healthcheck/bedrock.rb
```diff
@@ -0,0 +1,18 @@
...
+    attr_reader :message
+
+    def name
+      :bedrock
+    end
+
+    def status
+      models = Aws::Bedrock::Client.new.list_foundation_models.model_summaries.map(&:model_id)
+
+      models.any? ? GovukHealthcheck::OK : GovukHealthcheck::CRITICAL
```
kevindew, line 12: Can we check that we have the models we use? perhaps there is a method we can add to our BedrockModels module for it?

Logic behind this is that when we want to use a new model we need to enable this manually with clickops in each environment and have basically no way to verify that this has been done with terraform. So this healthcheck would at least provide that information. For that end it'd also be useful that this provided some text if a model wasn't enabled so it could be debugged.

lib/bedrock_models.rb
```diff
@@ -1,6 +1,8 @@
 module BedrockModels
-  CLAUDE_SONNET = ENV.fetch("CLAUDE_SONNET_MODEL_ID", "eu.anthropic.claude-sonnet-4-20250514-v1:0").freeze
-  TITAN_EMBED_V2 = "amazon.titan-embed-text-v2:0".freeze
+  MODELS = {
```
jackbot, line 2: A bit annoying to change this to a hash instead of using straight constants, but it made the healthcheck much easier to write. 

We could have done something like looping through `BedrockModels.constants` to get a list of all the models used, but it's a bit of a brittle way to do things and would cause the healthcheck to break if we added a constant that wasn't a model name.

kevindew:
Thanks @jackbot I think you may have got the wrong end of the stick (or I've misunderstood what you've demo'd to me).

I was wondering whether a new Healthcheck::Bedrock object was created each time a healthcheck was run (thus any instance variables are reset) or whether they cached the objects and ran status on them each time (expectation that each time status is run it does a check, so any cached data on it would invalidate the check).

In your example @message is changed each time status is called so wouldn't differentiate those scenarios (again unless I've missed something).

It seems a bit opaque as none of the Healthchecks in govuk_app_config store state in instance variables but I think [looking at the code it stores the objects once for the lifetime of the process](https://github.com/alphagov/govuk_app_config/blob/89162a857e488de5fd5ede80db849f9143e6f3b1/lib/govuk_app_config/govuk_healthcheck/checkup.rb#L25-L33), so any use of instance vars is dangerous

kevindew:
Ah yeah does look like we can rely on new objects created each time: 

https://github.com/alphagov/govuk_app_config/blob/main/lib/govuk_app_config/govuk_healthcheck.rb#L22-L24

so ivars seem safe afterall - sorry!

PR #443, opened 2025-08-06
WIP

Sunny-xyz:


Sunny-xyz:
image

We’ve updated the header examples to show the new limits (shown above, under the 'type' column), but if we want to call out rate-limits more prominently in the docs, we could add a dedicated Rate limits section at the top of our OpenAPI spec—like in the screenshot below

image

Sunny-xyz:
Thanks for the review Kev - I'll get on that now!

Sunny-xyz:
Rate limits have now been moved into their own section at the top of the docs for immediate visibility, shown below:

image

kevindew (approved):
Thanks looks good. I'll make a note on my todo list to update the description later

PR #449, opened 2025-08-08
Update the OpenAPI description

kevindew:
The talk of initial proof-of-concept seemed very out dated given how it is now an API implemented in the app.


davidgisbey (approved):
looks good 👍

PR #468, opened 2025-08-19
Fix error message display bug in grade C browsers

JamesCGDS:
## What
We're using the `replaceChildren()` function to display error messages, however this JS isn't compatible with grade C browsers (which should support JS enhancements). Therefore, an alternative approach has been implemented that works in these browsers.

Tested in Chrome 61, Edge 16 and 80, Firefox 60, and Safari 11 and everything seems to work okay.

## Why
[Trello card](https://trello.com/c/H7fBI1tX/2701-bug-error-messages-not-rendering-in-grade-c-browsers)

## Visual changes
Error messages now display in grade C browsers.

### DIFF
--- a/app/assets/javascripts/components/question-form.js
+++ b/app/assets/javascripts/components/question-form.js
@@ -129,7 +129,11 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         return li
       })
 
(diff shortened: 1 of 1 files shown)


kevindew (approved):
Change looks good. Bit of a shame about going for the older syntax for such niche browsers.

PR #481, opened 2025-08-21
Remove chat history message on homepage

JamesCGDS:
## What
Removes the chat history message on the homepage.

## Why
As per UCD discussion.

## Visual changes
The chat history message is no longer displayed on the homepage.


kevindew (approved):
Nice one

PR #482, opened 2025-08-21
Normalise newlines in user questions

davidgisbey:
## Description 

Now we're using a textarea and consuming questions via our API users can add newlines to their questions.

This has presented an issue where the character count component gets out of sync with the submitted character count.

I was able to replicate this on chrome by using shift enter to add a newline, which added a carriage return character before each "\n".

Because these are counted during the character length validation in the form, it would cause the validation to fail when the character count component indicated they were within the limit.

This updates the CreateQuestion form to normalise newlines before validation by removing carriage return characters.

This won't affect rendering, but it will ensure that the character count validation works as expected for users.

## Trello card

https://trello.com/c/egCTZWN5/2699-deal-with-problems-now-we-can-have-new-lines-in-chat-messages


kevindew (approved):
Looks good, I assume we use this same form for both API and web?

PR #483, opened 2025-08-21
Render question message using Govspeak component

davidgisbey:
## Description 

On the question show page on the Admin UI we probably want to capture the format the user submitted their question in. Including newlines.

This updates the row to use the Govspeak component to render the question message and updates the AnswersHelper to MessagesHelper to reflect that it's now used for both questions and answers.

## Screenshot

image

## Trello card

https://trello.com/c/egCTZWN5/2699-deal-with-problems-now-we-can-have-new-lines-in-chat-messages

### DIFF
--- a/app/helpers/admin/questions_helper.rb
+++ b/app/helpers/admin/questions_helper.rb
@@ -58,7 +58,7 @@ module Admin
         },
         {
           field: "Question",
-          value: question.message,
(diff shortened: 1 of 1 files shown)


davidgisbey:
ah i see. Thanks will sort that now

kevindew (approved):

PR #492, opened 2025-08-26
Bump API to V1

jackbot:
See individual commits for details, but in general this was a two-step process:

* Rename v0 to v1 for API routes (URLs, controllers, specs, etc)
* Add support for `/api/v0/*` URLs so that the old API continues to work while the app team upgrades the clients


davidgisbey (approved):
Looks great. I haven't read the v0 tests exhaustively since they're essentially a copy paste, but i did scan all the routes to check they're v0.

PR #495, opened 2025-08-27
Remove API v0

jackbot:
We recently re-versioned the API from v0 to v1, but we kept all the old `/api/v0/*` routes available while the backend and mobile devs on the app team moved everything over.

It's been confirmed that all that work has been done now, so we can safely remove the old v0 routes, and bump the major version for the API spec.


kevindew (approved):
Nice one. Looks good to me.

PR #498, opened 2025-08-27
Move shared .exportable functionality into shared example

davidgisbey:
## Description

We have the concept on .exportable on various models that we send to BigQuery. These largely have similar implementations. They grab records between two datetimes. This functionality can be extracted into a shared example which we can reuse.

The exception to this is the 'Question' model, which has additional logic for:

- preloading the conversation, answer and sources assoicated with it which we want to test
- it uses the answers created_at instead of the questions created_at

After thinking for a while about whether i should use the shared example for questions due to these differences I've chosen to use the shared example for shared functionality. I've left the tests that test additional functionality in the questions .exportable describe block.

I'm not wed to this though as it's a bit harder to parse what's actually going on than leaving the tests as is. So if anyone disagrees i'm happy to go with that.

## Trello card

https://trello.com/c/Lj6yUvvm/2748-add-additional-system-specs-to-check-user-interactions-are-displayed-in-admin-area

### DIFF
--- a/spec/models/answer_analysis_spec.rb
+++ b/spec/models/answer_analysis_spec.rb
@@ -4,26 +4,14 @@ RSpec.describe AnswerAnalysis do
   end
 
   describe ".exportable" do
-    let!(:new_answer_analysis) { create(:answer_analysis, created_at: 2.days.ago) }
+    it_behaves_like "exportable by start and end date" do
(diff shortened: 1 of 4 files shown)


davidgisbey:
@kevindew thanks i've made those changes

kevindew (approved):
Nice one

PR #511, opened 2025-08-29
API: Require end user ID header

jackbot:
https://trello.com/c/LUgo03zy/2733

This header value is currently optional. If it's set then it'll
associate a conversation with the user ID specified. If it's not, then
the conversation is just associated with the signon user.

One DPIA requirement for Chat is that a user can request deletion of
their data. Without requiring this header value we wouldn't be able to
find the conversations for a given user. So we need to require it.

### DIFF
--- a/docs/api_openapi_specification.yml
+++ b/docs/api_openapi_specification.yml
@@ -440,7 +440,7 @@ components:
     EndUserIdHeader:
       name: Govuk-Chat-End-User-Id
       in: header
-      required: false
+      required: true
       description: |
         An identifier for an individual end-user client to be used to provide
         individual end-user rate limiting to ensure that no one client can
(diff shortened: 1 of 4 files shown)


kevindew (commented):
This looks good to me, sorry I didn't submit the review yesterday I just left it with a pending comment 🤦 

The part of the card that doesn't seem to be covered here is the risk of an empty string being provided. Can we add something to the schema that asserts it's a string with some alphanumeric characters (perhaps \w is appropriate in a regex?)

docs/api_openapi_specification.yml
```diff
@@ -451,6 +451,7 @@ components:
         header, it's up to the client to ensure that the value is unique.
       schema:
         type: string
+        pattern: '^\S(.*\S)?$' # anything but an empty or all whitespace string
```
kevindew, line 454: I wonder if this is a bit more complex than it needs to be as I don't think we need to assert the whole string or that the first character is a non space (since we run strip on it). You might find just '\S' is enough to assert that it contains at least one non space character

kevindew (approved):
Nice one

PR #517, opened 2025-09-03
Add evaluation:topics_for_question rake task

davidgisbey:
## Description 

We want to add auto eval of topics to the evaluation repo. In order to do that we need to add a rake task to chat that returns the topics for a user question/input.

## Trello card 

https://trello.com/c/jxwjoYlu/2710-auto-evaluator-for-the-topic-tagger


lib/tasks/evaluation.rake
```diff
@@ -152,4 +152,18 @@ namespace :evaluation do
...
+    raise "Requires an INPUT env var" if ENV["INPUT"].blank?
+
+    result = AnswerAnalysisGeneration::TopicTagger.call(ENV["INPUT"])
+
+    output = {
+      primary_topic: result.primary_topic,
+      secondary_topic: result.secondary_topic,
+    }
+
+    puts(output.to_json)
```
kevindew, line 167:
```suggestion
    puts(result.to_json)
```

I imagine we could skip the intermediary step

spec/lib/tasks/evaluation_spec.rb
```diff
@@ -244,10 +251,7 @@
 
     before { Rake::Task[task_name].reenable }
 
-    it "requires an INPUT env var" do
-      expect { Rake::Task[task_name].invoke("openai") }
-        .to raise_error("Requires an INPUT env var")
-    end
+    it_behaves_like "a task requiring an input"
```
kevindew, line 254: Nice, I imagine if we find future use we could refactor this to `it_behaves_like "a task requiring an env var", "INPUT"` to check other env vars

kevindew (approved):
Looks good, just a quick suggestion on the JSON but doubt it needs another review.

lib/tasks/evaluation.rake
davidgisbey: Do we want to send back the llm_response and metric too though? That's quite a lot of noise.

kevindew: I'd have thought that'd be fine - the caller doesn't have to use them.

PR #521, opened 2025-09-03
Run onboarding animation only once

JamesCGDS:
## What
This PR prevents the onboarding animation from running again if the user clears their chat or refreshes the page without having started a conversation. This has been achieved by adding a new `govuk_chat_onboarding_complete` cookie to detect whether the onboarding animation has completed. Happy to discuss.

If we're happy with this approach, we'll need to merge a corresponding PR in the components gem before merging this to accept the `govuk_chat_onboarding_complete` cookie in order for this to work as we're using `GOVUK.cookie()` (hence the failing test).

## Why
Requested by design.

## Visual changes
N/A - no changes have been made to the animation itself, only when it is run.


kevindew (approved):
Looks good to me 👍

PR #544, opened 2025-09-15
Exclude opted out users from big query

davidgisbey:
## Description

App users may want their data excluded from analysis. I've opened a draft PR to add some config to our private repo which will contain an array of user ids to exclude https://github.com/alphagov/govuk_chat_private/pull/41 

This PR does the following

- ensures we exclude data for opted-out users from Biguery by updating the various `.exportable` scopes to filter out records associated with end users that have opted out
- add a rake task to delete records from BigQuery associated with an end user id

## Trello card

https://trello.com/c/ixnHciM4/2767-process-to-delete-data-from-big-query-if-someone-has-opted-out

### DIFF
--- a/app/models/answer_analysis.rb
+++ b/app/models/answer_analysis.rb
@@ -4,7 +4,9 @@ class AnswerAnalysis < ApplicationRecord
   belongs_to :answer
 
   scope :exportable, lambda { |start_date, end_date|
-    where(created_at: start_date...end_date)
+    joins(answer: { question: :conversation })
+    .where(created_at: start_date...end_date)
+    .merge(Conversation.exclude_opted_out_end_user_ids)
   }
 
   def serialize_for_export
--- a/app/models/conversation.rb
+++ b/app/models/conversation.rb
@@ -5,6 +5,13 @@ class Conversation < ApplicationRecord
 
   scope :active, -> { where(Question.active.where("questions.conversation_id = conversations.id").arel.exists) }
 
+  scope :exclude_opted_out_end_user_ids, lambda {
+    ids = Rails.configuration.govuk_chat_private&.opted_out_end_user_ids || []
+
+    where(end_user_id: nil)
+    .or(where.not(end_user_id: ids))
+  }
+
   enum :source,
        {
          api: "api",
--- a/app/models/conversation.rb
+++ b/app/models/conversation.rb
@@ -19,6 +19,16 @@ class Conversation < ApplicationRecord
        },
        prefix: true
 
+  def self.hashed_end_user_id(end_user_id)
+    return nil if end_user_id.blank?
+
+    OpenSSL::HMAC.hexdigest(
+      "SHA256",
+      Rails.application.secret_key_base,
+      end_user_id,
+    )
+  end
+
   def questions_for_showing_conversation(only_answered: false, before_id: nil, after_id: nil, limit: nil)
     scope = Question.where(conversation: self)
                   .includes(answer: %i[feedback sources])
@@ -55,10 +65,6 @@ class Conversation < ApplicationRecord
   def hashed_end_user_id
     return nil if end_user_id.blank?
 
-    OpenSSL::HMAC.hexdigest(
-      "SHA256",
-      Rails.application.secret_key_base,
-      end_user_id,
-    )
+    self.class.hashed_end_user_id(end_user_id)
   end
 end
--- a/lib/tasks/bigquery.rake
+++ b/lib/tasks/bigquery.rake
@@ -51,4 +51,29 @@ namespace :bigquery do
 
     puts "Deleted all #{records} BigqueryExport records"
   end
+
+  desc "Delete all BigQuery exports for an opted-out user"
+  task :delete_opted_out_end_user_data, %i[end_user_id] => :environment do |_, args|
+    end_user_id = args[:end_user_id]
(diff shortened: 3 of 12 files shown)


app/models/answer_analysis.rb
```diff
@@ -5,6 +5,15 @@ class AnswerAnalysis < ApplicationRecord
...
     where(created_at: start_date...end_date)
+    .where(
+      answer_id: Answer.select(:id)
+                       .joins(:question)
+                       .where(
+                         questions: {
+                           conversation_id: Conversation.select(:id).exclude_opted_out_users,
+                         },
+                       ),
+    )
```
kevindew, line 16: I don't think it's actually necessary to remove analysis and feedback since they don't contain personal data, but doesn't necessarily seem a problem. 

Could we move this to be a scope on answer perhaps? 

```
Answer.exclude_opted_out_end_user_ids.select(:id) 
```

I assume this makes this a subquery rather than executing a query first btw? (which is very cool)

app/models/conversation.rb
```diff
@@ -5,6 +5,16 @@ class Conversation < ApplicationRecord
 
   scope :active, -> { where(Question.active.where("questions.conversation_id = conversations.id").arel.exists) }
 
+  scope :exclude_opted_out_users, lambda {
+    excluded_ids = Rails.configuration.govuk_chat_private&.opted_out_user_ids || []
```
kevindew, line 9:
```suggestion
  scope :exclude_opted_out_end_user_ids, lambda {
    excluded_ids = Rails.configuration.govuk_chat_private&.opted_out_end_user_ids || []
```

Just a couple of minor naming suggestions for consistency

app/models/conversation.rb
```diff
@@ -5,6 +5,16 @@ class Conversation < ApplicationRecord
...
+
+    where(
+      arel_table[:end_user_id]
+        .not_in(excluded_ids)
+        .or(arel_table[:end_user_id].eq(nil)),
```
kevindew, line 14: I don't think we want the or here, if we don't have an end_user_id we plan to export the data

lib/tasks/bigquery.rake
```diff
@@ -51,4 +51,40 @@ namespace :bigquery do
 
     puts "Deleted all #{records} BigqueryExport records"
   end
+
+  desc "Delete all BigQuery exports for an opted-out user"
+  task :delete_opted_out_user_data, %i[end_user_id] => :environment do |_, args|
```
kevindew, line 56: I'm not too sure about using our own data here to work out what to delete - seems a bit complicated and requires the two datasets to be in sync, I think things could be simpler if we instead run `DELETE * from questions WHERE end_user_id = x` 

for x we'd have to obfuscate the value as per https://github.com/alphagov/govuk-chat/blob/1fe0ddae9dd55f83082a904dd005ef252c4bf7b0/app/models/conversation.rb#L48-L56

I don't think we want to worry about the feedback and analysis.

Happy to chat about this

kevindew (commented):
Looks great - I do wonder about seriously simplifying the big query delete, let me know if you'd like to chat. I've not reviewed in loads of detail at this point since that's probably a discussion point.

app/models/answer_analysis.rb
davidgisbey: so i wonder if it would make more sense to have it on a question since that's where it would be being used

```
  scope :exportable, lambda { |start_date, end_date|
                       joins(:conversation, :answer)
                       .preload(:conversation, answer: %i[sources])
                       .where("answer.created_at": start_date...end_date)
                       . exclude_opted_out_end_user_ids
                     }

```

lib/tasks/bigquery.rake
kevindew: Yeah, not identifiable so fine to remain.

app/models/conversation.rb
davidgisbey: Does 

```
 scope :exportable, lambda { |start_date, end_date|
                       excluded_ids = Rails.configuration.govuk_chat_private&.opted_out_user_ids || []

                       joins(:conversation, :answer)
                       .preload(:conversation, answer: %i[sources])
                       .where("answer.created_at": start_date...end_date)
                       .where.not("conversations.end_user_id": excluded_ids)
                     }
```

include nils? I thought it stripped them out which is why i added that or query?

davidgisbey:
Thanks Kev. I'll make that change. The private repo PR is here https://github.com/alphagov/govuk_chat_private/pull/41

lib/tasks/bigquery.rake
```diff
@@ -51,4 +51,29 @@ namespace :bigquery do
...
+    abort "You must provide an end_user_id" if end_user_id.blank?
+
+    bigquery = Google::Cloud::Bigquery.new
+    project_id = bigquery.project_id
+    dataset_id = bigquery.dataset(Rails.configuration.bigquery_dataset_id).dataset_id
+    hashed_end_user_id = OpenSSL::HMAC.hexdigest(
+      "SHA256",
+      Rails.application.secret_key_base,
+      end_user_id,
+    )
```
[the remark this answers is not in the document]
davidgisbey: Thanks i've done that 👍

kevindew:
@davidgisbey I think you might not have pushed up that commit.

davidgisbey:
@kevindew sorry i was just finishing writing it. Should've waited to ping you until it was done. It's in the latest commit

PR #546, opened 2025-09-16
Stop marking documents as ingested when indexing was skipped

chaecramb:
This PR facilitates work to present the last time we successfully synchronised a piece of content from the message queue in Grafana, and then to setup alerts if nothing has landed for days. Trello: https://trello.com/c/cUS4T1mU/2772-prometheus-guage-and-graph-for-last-document-ingested

Previously we bumped the gauge even when a payload was skipped, which would make the dashboard look healthy despite no indexing happening. This change distinguishes between a skipped change and a real one so that we only record when content is actually indexed.


kevindew (approved):
Looks good 👍

PR #548, opened 2025-09-16
Store AnswerSourceChunk records when writing answer sources

kevindew:
Trello: https://trello.com/c/qAuq0tcr/2782-store-used-search-results-in-the-chat-db

This builds on the work in https://github.com/alphagov/govuk-chat/pull/540 to start populating the AnswerSourceChunk table when sources are written to the database.

A change of note is that in the process of Answer creation we will write AnswerSourceChunks to the database before an answer is created (while we're still building unpersisted Ruby objects). This has been done to minimise the risk that we hit a unique index conflict by looking a record up and then having another written in the time before we write to the database. It increases a risk that we could have orphan AnswerSourceChunks in our database, but we risk this anyway with data retention so can hopefully clean them at the same time.

The next steps for this work are:

- backfill existing data so the new record can be relied upon
- switch to using the new data fields
- delete the old data

### DIFF
--- a/app/models/answer_source.rb
+++ b/app/models/answer_source.rb
@@ -13,6 +13,6 @@ class AnswerSource < ApplicationRecord
   end
 
   def serialize_for_export
-    as_json
+    as_json(except: :answer_source_chunk_id)
   end
 end
--- a/spec/models/answer_source_spec.rb
+++ b/spec/models/answer_source_spec.rb
@@ -25,9 +25,10 @@ RSpec.describe AnswerSource do
   end
 
   describe "#serialize for export" do
-    it "returns a source serialzed as json" do
-      source = create(:answer_feedback)
-      expect(source.serialize_for_export).to eq(source.as_json)
+    it "returns a source serialzed as json without the answer_source_chunk_id" do
+      source = build(:answer_source)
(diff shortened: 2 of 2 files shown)


app/models/answer_source_chunk.rb
```diff
@@ -1,2 +1,23 @@
 class AnswerSourceChunk < ApplicationRecord
+  def self.find_or_create_from_search_result(search_result)
```
davidgisbey, line 2: very nice 👌 Thanks for the link in the commit.

davidgisbey (approved):
should [these](https://github.com/alphagov/govuk-chat/blob/19db6da80ac627cf151aa548c41715b6e602b902/spec/lib/tasks/evaluation_spec.rb#L394-L412) also be weighted?

Other than that looks good. I don't think it needs a re-review so i'll ✅

kevindew:
> should [these](https://github.com/alphagov/govuk-chat/blob/19db6da80ac627cf151aa548c41715b6e602b902/spec/lib/tasks/evaluation_spec.rb#L394-L412) also be weighted?
> 
> Other than that looks good. I don't think it needs a re-review so i'll ✅

Oh yes, I think they can be made a bit nicer with the new factory. Will update.

PR #557, opened 2025-09-22
Add completeness graph to metrics & reorder them

davidgisbey:
## Description

This adds an answer completeness graph to the admin metrics. I've utilised the same logic as the answer feedback graph. So grouped by hour for 24hrs and grouped by day for 7 days. Happy to just have a simple pie chat for per day if that's preferable.

I've also reordered the metrics. The reasoning for the new ordering is in the commit. Again v happy to discuss/tweak based on peoples feedback.

### Last 24 hours 

image

### Last 7 days

image

### New ordering 

image

## Trello card

https://trello.com/c/UsyXUPCd/2805-add-answer-completeness-graph-to-the-admin-area-metrics

### DIFF
--- a/app/controllers/admin/metrics_controller.rb
+++ b/app/controllers/admin/metrics_controller.rb
@@ -18,12 +18,6 @@ class Admin::MetricsController < Admin::BaseController
     render json: count_by_period(scope, :created_at).chart_json
   end
 
-  def answer_feedback
-    scope = AnswerFeedback.where(created_at: start_time..)
-                          .group_useful_by_label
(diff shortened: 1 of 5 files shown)


kevindew (commented):
Looks good. 

I do imagine going pie chart might be better for completeness as I can't imagine enough happens over an hour to make it be worthwhile by hour - but really don't mind too much.

Was going to suggest killing feedback as per it's lack of role in API but after reading your commit message just having at the bottom is fine too. Would probably suggest removing if you re-edit this but if you don't need to further it's fine where it is.

kevindew (approved):
nice one

PR #566, opened 2025-09-25
Create error_timeout answers when appropriate in the GET answer API endpoint

davidgisbey:
## Description 

We have the concept of an answer timing out in the Chat UI. When an answer takes longer than 2 minutes to generate, we create an "error_timeout" answer so that the user is able to ask another question.

We didn't add this to the API endpoint, so this commit adds that functionality.

## Trello card

https://trello.com/c/QG0VO4sA/2812-update-api-to-create-timed-out-answer-after-2-minutes


kevindew (approved):
Looks good 👍

PR #582, opened 2025-09-30
Remove exception messages from healthchecks

jackbot:
Our healthchecks are accessible on the public internet and currently
we're returning messages from them which expose certain details about
our application and infrastructure.

While these messages are useful to us for debugging, it's probably not
wise to expose them publicly.

Platform Engineering have confirmed that there's no precedent for these
routes not to be public, so it seems the best option is to remove all of
the custom messages we return and simply just return the `status`
attribute.


kevindew (commented):
I think we can be more lenient with what we expose here - it seems fine for us to expose pre-prepared messages but more delicate when we have next to no control over what it could be.

PR #583, opened 2025-09-30
Add non_llm_answer to AnswerStrategy enum

davidgisbey:
## Description

This strategy has been added to a branch used for load testing. When that branch is deployed and answers are generated, we persist the answer with the non_llm_answer strategy.

When we redeploy out main, it returns nil since the value isn't in our enum. This commit adds the value to the enum so that we can read those records and the question show page in the admin UI doesn't blow up when it tries to humanise nil.


kevindew (approved):
Looks good. I did wonder if we're at risk of this list being exposed anywhere but surely we're not given it has the old OpenAI RAG Completion one.

davidgisbey:
Yea. Once we wipe the db prior to the experiment i wonder if we should remove both?

PR #587, opened 2025-10-03
Fix pagination bugs on questions index page 

davidgisbey:
## Description 

When we added a bunch of new filters, we missed adding them to the pagination logic, which means that the query params aren't retained in the pagination links.

This adds them to the pagination_query_params method in the QuestionsFilter class.

I've also done a bit of work to pull out some shared creation logic to it's own method and updated the test to build the expected_params from the filter attrs.

Hopefully this way it'll mean that if any new filters are added via attrs and not added to the filter params logic, the filter param retention tests will fail.

## Trello card

https://trello.com/c/EH4JTpYO/2825-fix-admin-area-user-pagination-bug

### DIFF
--- a/app/models/admin/filters/questions_filter.rb
+++ b/app/models/admin/filters/questions_filter.rb
@@ -69,11 +69,18 @@ private
     filters = {}
     filters[:status] = status if status.present?
     filters[:search] = search if search.present?
+    filters[:source] = source if source.present?
     filters[:start_date_params] = start_date_params if start_date_params.values.any?(&:present?)
     filters[:end_date_params] = end_date_params if end_date_params.values.any?(&:present?)
     filters[:sort] = sort if sort != self.class.default_sort
     filters[:answer_feedback_useful] = answer_feedback_useful unless answer_feedback_useful.nil?
     filters[:conversation_id] = conversation.id if conversation.present?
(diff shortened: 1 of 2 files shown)


kevindew (commented):
This looks great, thanks for fixing. 

Given we missed so many do we want to consider refactoring the tests for questions_filter so that when we test whether a filter works it also tests the url? Feels potentially like a shared context

davidgisbey:
We def can. I've updated the [tests](https://github.com/alphagov/govuk-chat/pull/587/files#diff-419186c758d0e274f0b8a0e6ef6538e41ddb9c2966f832673c0dca0ff3db0d23R372) to use the filter's attributes to build the expected result. 

My thinking is that is someone adds a new attribute to the filter class and forgets to update the [pagiation_query_params](https://github.com/alphagov/govuk-chat/blob/9e3d1562fcbe00b8d46578eb9628d5bd2db83b3a/app/models/admin/filters/questions_filter.rb#L68-L78) then the pagination tests will fail.

But maybe that's a bit of a pain and slightly brittle.

kevindew:
Gotcha, yeah that sounds good it meets the need that I'd have that an additional field would provide a failing test.

kevindew (approved):

PR #622, opened 2025-10-20
[DO NOT MERGE][FOR DEMO ONLY] switch Chat to use Haiku 4.5

exfalsoquodlibet:
ad-hoc temporary change to switch Chat to use Haiku 4.5 for demo purposes only


kevindew:
This is old and unlikely to be picked up again

PR #1002, opened 2026-04-13
[CHAT-472] Use Claude as default LLM provider for answer composition

davidgisbey:
## Description

We're going to remove our OpenAI integration. This is a bit of preliminary work to make that easier.

We've got some guardrail classes that support both OpenAI and Claude, but default to OpenAI. This PR updates these to default to Claude instead.

## Jira ticket 


kevindew (approved):
Looks good

PR #1004, opened 2026-04-13
[CHAT-472] Update rake tasks to not take a provider argument

davidgisbey:
## Description

This is one of many PRs that will remove our OpenAI integration. Since we're only going to be supporting Claude going forward we can remove the provider argument from the evaluation rake tasks and call the Claude pipeline directly.

I've also opened a [PR](https://github.com/alphagov/govuk-chat-evaluation/pull/278) on the evaluation repo which removes the provider configuration. That PR should be merged shortly after this one.

## Jira ticket

https://gdsgovukagents.atlassian.net/jira/software/c/projects/CHAT/boards/269?label=Backend_Dev&selectedIssue=CHAT-472

PR #1005, opened 2026-04-13
[CHAT-472] Remove OpenAI answer composition integration

davidgisbey:
## Description

This PR follows on from #1004 which removed the provider argument from the evaluation rake tasks. This PR focusses on removing our OpenAI answer composition integration. It:

- removes the logic related to the`openai_structured_answer` answer strategy from the composer
- updates the pipeline runner to remove OpenAI specific error handling
- marks the `openai_structured_answer` strategy as deprecated
- removes logic related to the `openai_structured_answer` from shared components (guardrails etc.)
- removes the OpenAI pipeline steps
- removes the OpenAI e2e system spec

It should be noted that there will be follow up work to remove the Claude namespace for our pipeline components, but this will be handled in a separate PR to keep the scope of this change focused on removing OpenAI.

## Jira ticket 


lib/answer_composition/pipeline/question_rephraser.rb
```diff
@@ -14,8 +14,6 @@ def call(context)
 
         start_time = Clock.monotonic_time
         klass = case llm_provider
```
kevindew, line 16: Should we also be removing this argument?

lib/guardrails/jailbreak_checker.rb
```diff
@@ -57,8 +57,6 @@ def initialize(input, llm_provider = :claude)
 
     def call
       case llm_provider
```
kevindew, line 59: As above, should we be removing this arg?

lib/guardrails/multiple_checker.rb
```diff
@@ -93,8 +93,6 @@ def initialize(input, llm_prompt_name, llm_provider)
 
     def call
       case llm_provider
```
kevindew, line 95: repeat comment

kevindew (commented):
Looks good, but a few comments about the llm_provider variable

lib/answer_composition/pipeline/question_rephraser.rb
davidgisbey: I can do it as part of this, but my plan was to have a follow up PR that merges the classes. This would be done as part of that.

kevindew: Great no worries to be a follow up PR, I just couldn't see it in the existing follow up PR. I'll approve this.

kevindew (approved):

PR #1006, opened 2026-04-13
[CHAT-472] Remove remaining OpenAI integration

davidgisbey:
## Description

This follows on from #1005 which remove the OpenAI answer composition integration. This PR remove the rest of our OpenAI integration. It:

- Removes the unused `Search::TextToEmbedding::OpenAI` class.
- Removes the OpenAI healthcheck
- Removes the OpenAIClient and related application config
- Removes our OpenAI prometheus metrics
- Remove the OpenAI stubs
- Removes Heroku config
- Updates references to OpenAI in the config to Anthropic
- Remove the `ruby-openai` gem

## Jira ticket 


docs/guardrails.md
```diff
@@ -34,7 +34,7 @@ The output of the LLM is as follows:
 
 We map these to meaningful names using the mappings from a config file, e.g. [here](../config/llm_prompts/answer_guardrails.yml).
 
-The file also contains the prompts we use to run the guardrails. Copy/paste these into the [OpenAI chat playground](https://platform.openai.com/playground/chat?models=gpt-4o) to investigate any issues.
+The file also contains the prompts we use to run the guardrails. Copy/paste these into the [Antrhopic workbench](https://platform.claude.com/workbench) to investigate any issues.
```
kevindew:
```suggestion
The file also contains the prompts we use to run the guardrails. Copy/paste these into the [Anthropic workbench](https://platform.claude.com/workbench) to investigate any issues.
```

davidgisbey: ah good spot. I've sorted that.

app/views/static/about.html.erb
kevindew, line 1: Nice one sorting this page. We probably need to work out what to do with it one day as it just sits there without any links to it :-/

app/models/question.rb
```diff
@@ -1,8 +1,6 @@
 class Question < ApplicationRecord
   enum :answer_strategy,
        {
-         open_ai_rag_completion: "open_ai_rag_completion", # legacy strategy - no longer used
-         openai_structured_answer: "openai_structured_answer", # legacy strategy - no longer used
```
kevindew, line 5: I read your PR comment, I think it's safe to delete them. It'd just be an admin area error - could well be worthing pinging the dev channel first

kevindew (approved):
Looks good 👍

app/views/static/about.html.erb
davidgisbey: Yea i did wonder about just removing it, but figured it wasn't doing any harm and we might want it for the web version (although it seems unlikely).

app/models/question.rb
davidgisbey: Great will do.

app/views/static/about.html.erb
kevindew: Removing it would cause some minor harm as it's still part of our e2e tests: https://github.com/alphagov/govuk-e2e-tests/blob/2f3f09e210cc8251144699a5028433b8772a18ae/tests/chat.spec.js#L5-L12 I imagine we may want a different test or even to just remove that one

PR #1015, opened 2026-04-14
[CHAT-472] Remove Claude pipeline steps and titan namespacing

davidgisbey:
## Description 

This is the first of 3 PRs that will attempt to merge some of the classes we've setup that handled multiple LLM providers being in place. Since we're now only using Claude, we can get rid of the Claude name spacing. This PR:

- moves the logic from the titan TextToEmbedding::Titan class into TextToEmbedding  since we only have 1 embedding provider available
- Merges the question rephraser classes into one class
- Moves the structured answer comp and question router out of the Claude dir into the pipeline steps

Follow up PRs will look at combining some of the jailbreak and output guardrails code.

## Jira ticket 


kevindew (approved):
Great job ⭐ 

This is a bit of a tough PR to review with it being nearly all code moving but the UI not really making clear what was existing code.

So I've approved it on expectation that skimming it, all looks sane but could you let me know if there is anything newly introduced that should be reviewed?

davidgisbey:
Everything is just a rename except the [question rephraser commit](https://github.com/alphagov/govuk-chat/pull/1015/changes/89afc789dfffadfb798b360216274c2a69e12b54). So nothing new was added 👍

kevindew:
Ace - thanks

PR #1019, opened 2026-04-15
[CHAT-472] Merge jailbreak guardrails into single pipeline class

davidgisbey:
## Description

We've got 3 classes for Jailbreak guardrails. It's a fairly simple class and can just be moved into a single class in the answer composition pipeline.

I did consider leaving the JailbreakChecker class as a separate class, but it just didn't seem like there was a lot of value.

One thing i noticed while merging this, is that we create a ResponseError class in the JailbreakChecker, but we don't actually raise it anywhere so it was redundant. I've removed it, but if we want we can check that the llm response either returns the pass or fail value and create the object in the db with the correct status if it doesn't.

We'd need to add some additional prompt config for the fail value though in order to do that so i've avoided it for now.

We will need to do some follow up work to update the evaluation repo since we just return the serialised answer now so it'll need to grab the info it wants from the serialised answer instead of the result that was previously returned from the JailbreakChecker.

## Jira ticket 


lib/answer_composition/pipeline/jailbreak_guardrails.rb
```diff
@@ -1,45 +1,96 @@
...
         context.answer.assign_metrics("jailbreak_guardrails", build_metrics(start_time, response))
 
-        if response.triggered
+        if jailbreak_guardrails_status == :fail
           context.abort_pipeline!(
             message: Answer::CannedResponses::JAILBREAK_GUARDRAILS_FAILED_MESSAGE,
             status: "guardrails_jailbreak",
           )
         end
-      rescue Guardrails::JailbreakChecker::ResponseError => e
```
davidgisbey, line 22: So we had this but it was never actually being used. I'm not sure if we want to remove it altogether. Or check to see if the response matches the success or fail value and if not raise the error.

This will also effect [this](https://github.com/alphagov/govuk-chat/pull/1019/changes#diff-ca44027c297f321c42ea605f5e29f16dc516da337fc2b01e69a328003d276cddL24-L30).

jackbot (approved):

lib/answer_composition/pipeline/jailbreak_guardrails.rb
davidgisbey: @kevindew no big rush, but it'd be good go get your thoughts on this.

Basically:

- the code is unused. There is no situation where this error is raised
- it looks like it's a hangover from the Multiple checker which
  - checks the llm response again a regex to see if it matches the pattern "true | 1,2" etc.
  - if not it raises a response error
- the evaluation rake task also rescues the error and builds a [bespoke JSON payload](https://github.com/alphagov/govuk-chat/pull/1019/changes#diff-ca44027c297f321c42ea605f5e29f16dc516da337fc2b01e69a328003d276cddL24-L30).

My thoughts are we've got 2 options:

1. remove it altogether and update the rake task to just return the serialised answer. The eval repo can grab what it needs from the payload 
2. add a `fail_value` to the jailbreak prompt config and add some logic to the jailbreak class to check it's either the pass or fail value. If not raise the error

We've gone a long time without any of this functionality actually working though so i'm not massively convinced we need it. It does mean we've got a bit of a dead status though since  `error_jailbreak_guardrails is only ever used in that rescue.

kevindew: Thanks for explaining. 

Hmm with this error being redundant I think we've still kinda get the same behaviour as if it was there but not the extra ease of filtering for it that we do with a dedicated status. I.e:

if we pass a guardrail, we continue with the pipeline
if we fail a guardrail, we abort with a guardrail failed status
if we get an unexpected llm response, we abort with a guardrail failed status

I think your second option sounds the best as we could achieve the above but with a dedicated status so that it would be easy to tell when we get a response we can't deal with - it feels somewhat useful to track that given we want to add structured responses to other guardrails to prevent odd responses.

davidgisbey: Oh yes sorry i said raise the error, but i meant abort like you said. I'll add that in as a separate commit.

davidgisbey:
@kevindew i've added handling an invalid LLM response in the last 2 commits.

lib/answer_composition/pipeline/jailbreak_guardrails.rb
```diff
@@ -1,45 +1,116 @@
...
+          messages:,
+          **inference_config,
+        )
+
+        text_response = response[:content][0][:text]
+        jailbreak_guardrails_status = if llm_text_response_invalid?(text_response)
+                                        :error
+                                      else
+                                        text_response == pass_value ? :pass : :fail
+                                      end
```
kevindew: Minor, I think you might find this is all a bit cleaner with a case statement e.g:

```
case text_response
when pass_value
  :pass
when fail_value
  :fail
else
  :error
end
```

kevindew (approved):
Looks good, just a minor code organisation comment

lib/answer_composition/pipeline/jailbreak_guardrails.rb
davidgisbey: Yea good shout that's way better 👍

PR #1020, opened 2026-04-15
[CHAT-472] Merge multiple checker into single class

davidgisbey:
## Description

I'll start this off saying i'm not 100% sure this is the right thing to do. I do feel like this class is probably doing a bit too much. It feels a bit painful to work with. I mention a potential refactor in the next section.

I do wonder if instead we should of this we should rename the dir to multiple checker and separate classes for the multiple checker and the prompt builder.

I do think that combining multiple checker logic is a better solution overall though. Even without the refactor.

I've also updated the OutputGuardrails and the classes that inherit from it to use instance methods instead of class methods so they're inline with the rest of the codebase and are easier to work with.

## Potential refactor 

I've opened this [PR](https://github.com/alphagov/govuk-chat/pull/1034) with a potential refactor. I wanted to see what people thought before dedicating longer than 10 mins to it though. 

## Jira ticket


kevindew (commented):
Nice job, this looks good. I'll now give some attention to the refactor PR - do agree that something should be done.

kevindew (approved):

PR #1022, opened 2026-04-15
Update QuestionRephraser supported models

davidgisbey:
## Description

We're actually going to be using Sonnet 4.5, not Haiku 4.5 for the question rephraser.

The prompt for Sonnet 4.5 was added in this [PR](https://github.com/alphagov/govuk_chat_private/pull/106):

This adds Sonnet 4.5 to`BedrockModels` and updates the `QuestionRephraser::SUPPORTED_MODELS` to include Sonnet 4.5 instead of Haiku 4.5.

### DIFF
--- a/spec/lib/healthcheck/bedrock_spec.rb
+++ b/spec/lib/healthcheck/bedrock_spec.rb
@@ -17,6 +17,7 @@ RSpec.describe Healthcheck::Bedrock do
 
       allow(BedrockModels).to receive(:MODEL_IDS).and_return({
         claude_sonnet_4_0: "eu.anthropic.claude-sonnet-4-20250514-v1:0",
+        claude_sonnet_4_5: "eu.anthropic.claude-sonnet-4-5-20250929-v1:0",
         claude_sonnet_4_6: "eu.anthropic.claude-sonnet-4-6",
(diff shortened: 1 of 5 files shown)


spec/lib/healthcheck/bedrock_spec.rb
```diff
@@ -70,7 +75,8 @@
         healthcheck.status
         expect(healthcheck.message).to eq(
           "Bedrock model(s) not available: amazon.titan-embed-text-v2:0, anthropic.claude-haiku-4-5-20251001-v1:0, " \
```
kevindew: Shall we change this test to just check the start of the string? it seems a bit unnecessarily painful to have to update this list each time we change model ids.

kevindew (commented):
Looks good, just a suggestion of a simplification.

spec/lib/healthcheck/bedrock_spec.rb
davidgisbey: Sounds good I've updated it to match on a single model name. I used `amazon.titan-embed-text-v2:0` since it seems like that's probably the least likely to be removed.

kevindew (approved):

PR #1027, opened 2026-04-16
Use auto-evaluation topic tagger prompt config for question topics

davidgisbey:
This was still using the Claude topic tagger config. Alessia realised the prompts hadn't been deleted and that this was still reliant on them despite us switching to the GPT OSS 120b model for topic tagging.


kevindew (approved):
Nice one

PR #1034, opened 2026-04-16
[CHAT-472] Move MultipleChecker into AnswerComposition namespace and break it into multiple classes

davidgisbey:
## Description

As part of pulling our OpenAI integration out of the codebase, we simplified a bunch of the code related to pipeline steps that were shared between our OpenAI and Bedrock implementations.

As part of this work, the Jailbreak Guardrails were moved into a single class in the answer composition pipeline directory. This has left us with the MultipleChecker being the only class in the Guardrails namespace.

Additionally, when I merged the MultipleChecker classes it was a bit of a painful process due to the class being quite large with multiple nested classes.

This PR does a few things: 

- It moves the 
- It breaks MultipleChecker out into 3 classes, one handles prompt construction, another defined the response error and the final class handles calling the LLM and parsing the response into a Result object.
- It removes the unused `.collated_prompts` method
- It updates the guardrails rake task to use the pipeline runner rather than calling the MultipleChecker so that it is in line with the rest of the rake tasks

On the latter point, we will need a follow up PR for the evaluation repo which grabs the info it needs from the serialised answer.

## Jira ticket 


lib/tasks/guardrails.rake
```diff
@@ -7,7 +7,7 @@ namespace "guardrails" do
       abort("Invalid guardrail type. Valid guardrail types are #{valid_guardrail_types.to_sentence}")
     end
 
-    prompt = Guardrails::MultipleChecker.collated_prompts(guardrail_type)
+    prompt = MultipleChecker::Prompt.collated(guardrail_type)
```
kevindew: I somewhat imagine we might not even need this rake task anymore?

kevindew (commented):
Thanks for digging into this David - looks a really good job and definitely an area to give attention.

There’s some foundational things that I wanted to see about exploring here that might lead to a neater design (well from my perspective).

I think the thing that feels strangest at the moment is that we had this guardrails namespace that just had one thing in it and I think if we get rid of that and have a MultipleChecker namespace we also have a bit of a confusing situation, so I wondered about what might make sense being moved into the AnswerComposition namespace like other parts.

First, I wondered: how painful would it be that the the evaluation rake task was to call an AnswerComposition class like our other evaluation rake tasks? As this seems to be the only place where the MultipleChecker class is exposed somewhat publicly compared to being almost a private dependency of the Guardrail classes.

If that could be resolved I’d wonder if it’d make sense to have the supporting classes within AnswerComposition. Could name it MultipleGuardrail and then have the Checker, Prompt and Response classes? This would feel a bit closer in placement to other supporting classes like LinkTokenMapper

I imagine all the logic would be the same, what do you think?

lib/tasks/guardrails.rake
davidgisbey: Yea i remember it being added, but i'm not sure who is using it. I'll check with DS and if not i'll pull it out.

davidgisbey:
@kevindew thanks for the feedback. That all sounds good and makes sense to me. I'll give that a go in this PR. If it drags on i'll add a new card, but i suspect it'll just a case of making the suggested changes and a small PR in the eval repo to handle a serialised answer being passed back. So it shouldn't be too bad.

davidgisbey:
@kevindew thanks for all the feedback. This one's ready for re-review.

spec/lib/answer_composition/multiple_guardrail/prompt_spec.rb
kevindew, line 1: Nit: you've named this prompts_spec.rb but it should be prompt_spec.rb

lib/answer_composition/pipeline/context.rb
```diff
@@ -4,7 +4,7 @@ class Context
 
     def initialize(question)
       @question = question
-      @answer = question.build_answer
+      @answer = question.answer || question.build_answer
```
kevindew, line 7: Your logic for doing this seems reasonable from the commit message, only concern I have is that it could trigger an exception if a question doesn't have an answer already because of the n+1 query prevention. Just thought I'd check whether there was any extra checks you felt would be appropriate for this if you hadn't clocked this previously?

Naturally, given it passes all the tests it should be fine otherwise we've got a coverage issue

kevindew (approved):
Looks great, just a couple of small things that hopefully don't need another review

lib/answer_composition/pipeline/context.rb
davidgisbey: So we have `has_one :answer, strict_loading: false` in the question model so i think we're fine. I could explicitly set it no nil in the ComposeAnswerJob to avoid the db hit though.

davidgisbey: aha we've already got it there

```
  def perform(question_id)
    question = Question.includes(:answer, :conversation).find_by(id: question_id)
    return logger.warn("No question found for #{question_id}") unless question
    return logger.warn("Question #{question_id} has already been answered") if question.answer

    answer = AnswerComposition::Composer.call(question)

    begin
      answer.save!
    rescue ActiveRecord::RecordNotUnique
      logger.warn("Already an answer created for #{question_id}")
    end

    AnswerAnalysis.enqueue_async_analysis(answer) if answer.persisted?
  end

```

So i think we're all good

PR #1044, opened 2026-04-17
adding sonnet 4.5 as a supported question_router model

ciaran-ryan-gds:
This adds Sonnet 4.5 to the supported models for the question_router.rb. 

This PR will be raised in tandem with adding Sonnet 4.5 to the govuk-chat-private repo where it refers directly to the config for Sonnet 4.0.

### DIFF
--- a/lib/answer_composition/pipeline/question_router.rb
+++ b/lib/answer_composition/pipeline/question_router.rb
@@ -1,6 +1,6 @@
 module AnswerComposition::Pipeline
   class QuestionRouter
-    SUPPORTED_MODELS = %i[claude_sonnet_4_0 claude_haiku_4_5 claude_sonnet_4_5].freeze
+    SUPPORTED_MODELS = %i[claude_sonnet_4_0 claude_sonnet_4_5].freeze
     DEFAULT_MODEL = :claude_sonnet_4_0
 
     def self.call(...) = new(...).call
--- a/lib/answer_composition/pipeline/question_router.rb
+++ b/lib/answer_composition/pipeline/question_router.rb
@@ -1,6 +1,6 @@
 module AnswerComposition::Pipeline
   class QuestionRouter
-    SUPPORTED_MODELS = %i[claude_sonnet_4_0 claude_haiku_4_5].freeze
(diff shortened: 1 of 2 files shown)


lib/answer_composition/pipeline/question_router.rb
```diff
@@ -1,6 +1,6 @@
 module AnswerComposition::Pipeline
   class QuestionRouter
-    SUPPORTED_MODELS = %i[claude_sonnet_4_0 claude_haiku_4_5].freeze
+    SUPPORTED_MODELS = %i[claude_sonnet_4_0 claude_haiku_4_5 claude_sonnet_4_5].freeze
```
davidgisbey: If there's no intention of using Haiku 4.5 shall we remove it? Other than that it Looks good 👍

ciaran-ryan-gds: @exfalsoquodlibet confirming removal of haiku as an option from question router. Should probably be removed from private repo too in this case.

ciaran-ryan-gds:
Removed Haiku 4.5 from the question router rb file supported models. If we are to change the private repo to delete the Haiku prompt from question router, will we need to change the git revision in the gemfile here again? @davidgisbey

ciaran-ryan-gds:
Note this test is failing due to the Haiku 4.5 existing in spec/lib/answer_composition/pipeline/question_router_spec.rb, "when using a model newer than Claude Sonnet 4", line 243. Should we change this to be claude_sonnet_4_5 also?

davidgisbey:
Thanks for doing that @ciaran-ryan-gds. It looks like Haiku 4.5 was being referenced in a test so i've just fixed that.

davidgisbey:
hmm a bunch of seemingly unrelated tests have started failing. Very bizarre. Let me have a look into it.

davidgisbey (approved):
Okay i've got that sorted. @ciaran-ryan-gds i'll merge this in if you're happy it's good to go.

ciaran-ryan-gds:
Yep looks great, thanks for fixing the tests @davidgisbey ! Happy to merge

PR #1050, opened 2026-04-20
[CHAT-473][DO NOT MERGE] Use structured output for non-sonnet 4.0 guardrail requests

davidgisbey:
## Description 

This moves the guardrail response from being a string to being a structured response. We want to keep non-structured responses in place for Sonnet 4, but all other models will use them.

There was some discussion about where the schema should live for the structured response. I initially put it in the private repo, but it's so simple that we've decided to add it here instead.

It does have a description which described the expected content of the response, but our code shows that anyway so I don't think it's a problem to have it here.

If we do care about it, then we can transition the response to use the guardrail names as the enum values and then map them using that instead of integers. But that changes the implementation quite a bit so i'm leaving it for now.

It might be something to look at once we've transitioned to using Haiku 4.5.

## Jira ticket

https://gdsgovukagents.atlassian.net/jira/software/c/projects/CHAT/boards/269/backlog?label=Backend_Dev&selectedIssue=CHAT-473

### DIFF
--- a/lib/answer_composition/multiple_guardrail/prompt.rb
+++ b/lib/answer_composition/multiple_guardrail/prompt.rb
@@ -16,7 +16,13 @@ module AnswerComposition::MultipleGuardrail
       guardrails_content = guardrails.map { |g| "#{g.key}. #{g.content}" }
                                      .join("\n")
 
-      prompts.fetch(:system_prompt)
+      system_prompt_key = if Checker.bedrock_model == :claude_sonnet_4_0
+                            :system_prompt
+                          else
+                            :system_prompt_structured
+                          end
+
+      prompts.fetch(system_prompt_key)
              .sub("{guardrails}", guardrails_content)
              .sub("{date}", Date.current.strftime("%A %d %B %Y"))
     end
--- a/spec/lib/answer_composition/multiple_guardrail/prompt_spec.rb
+++ b/spec/lib/answer_composition/multiple_guardrail/prompt_spec.rb
@@ -11,6 +11,16 @@ RSpec.describe AnswerComposition::MultipleGuardrail::Prompt do
       {guardrails}
     PROMPT
   end
+
+  let(:system_prompt_structured) do
+    <<~PROMPT
+      This is a structured system prompt and the date is {date}.:
+
+      The following guardrails must be followed:
+
+      {guardrails}
+    PROMPT
+  end
   let(:user_prompt) do
     <<~PROMPT
       This is the user prompt:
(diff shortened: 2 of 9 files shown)


kevindew (commented):
This looks good, I think my comments around this are just covered by the review on the prompt and how that may impact this: https://github.com/alphagov/govuk_chat_private/pull/114

Have you had a chance to try it out?

davidgisbey:
Thanks i'll have a look. Yea it worked fine when i tried it locally.

davidgisbey:
@kevindew this one is ready for review. I did leave a question on the other PR about whether we should just define the schema in this app since it's so simple.

lib/answer_composition/multiple_guardrail/prompt.rb
```diff
@@ -32,6 +32,17 @@ def guardrails
...
 
+    def json_schema
+      if Checker.bedrock_model == :claude_sonnet_4_0
+        raise NotImplementedError, "Structured responses are not supported for claude_sonnet_4_0"
+      end
+
+      schema = prompts.fetch(:json_schema)
+      enum_values = (1..guardrails.length).to_a
+      schema.fetch("schema").fetch("items")["enum"] = enum_values
+      schema
```
kevindew: Yeah as per comment on other PR, I agree it'd be much nicer to have whole schema.

davidgisbey:
I've added the JSON schema to this codebase as suggested. I've now worried about hiding the description because you can figure out what the keys are based on our code. If we want to update this to pass back a list of guardrail names instead of integers we can obfuscate the expected output format, but until then i don't think there's a lot of point.

spec/lib/answer_composition/multiple_guardrail/prompt_spec.rb
```diff
@@ -26,34 +36,69 @@
...
   end
+  let(:json_schema) do
+    {
+      "type" => "json_schema",
+      "schema" => {
+        "type" => "array",
+        "items" => { "type" => "integer" },
+      },
+    }
+  end
```
kevindew: I don't think this is needed anymore

spec/lib/answer_composition/multiple_guardrail/prompt_spec.rb
```diff
@@ -26,34 +36,69 @@
...
   let(:guardrails_config) do
     {
       system_prompt:,
+      system_prompt_structured:,
       user_prompt:,
       guardrails:,
       guardrail_definitions:,
+      json_schema:,
```
kevindew: I think this can be removed too

spec/support/stub_claude_messages.rb
```diff
@@ -163,10 +163,28 @@ def stub_claude_structured_answer(question_or_history,
...
+    if bedrock_model != :claude_sonnet_4_0
+      json_schema = {
+        "type" => "json_schema",
+        "schema" => {
+          "description" => "Array of triggered guardrail numbers. Returns [] if none triggered.",
+          "type" => "array",
+          "items" => { "type" => "integer", "enum" => guardrail_keys },
+        },
+      }
+      chat_options[:output_config] = { format: json_schema }
```
kevindew: This looks a bit coupled to code in checker, I'd wonder if we could just have `a_hash_including({ "type" => "json_schema" })`

kevindew (commented):
Looks great, just a few small things

spec/lib/answer_composition/multiple_guardrail/prompt_spec.rb
davidgisbey: Oh yes sorry good spot

spec/support/stub_claude_messages.rb
davidgisbey: Yea okay makes sense to me. I'll sort that 👍

kevindew (approved):
I see you've got a linting error but otherwise all looks good to me 👍

davidgisbey:
oh yes thanks. I'll confirm whether DS want to make any changes to the prompt at standup. If they don't i'll drop merge the private repo PR and drop the first commit. But we're good to any time after that.

PR #1059, opened 2026-04-22
Switch default models for Guardrails to Haiku 4.5

kevindew:
This flips the switch of which models we'll be using in the codebase for these guardrail tasks and represents the first stage in a multi-step journey to change the models.

Following the adoption of this model as the default we will later remove the Sonnet 4.0 configuration - once we're content we don't need to migrate back.

In order to get the tests passing, some changes were required to the `StubClaudeMessages` helpers. The `claude_messages_response` hardcoded the Sonnet 4.0 model in the response. I've updated the helper to allow an optional `bedrock_model` to be passed in and used to populate it instead.

### DIFF
--- a/lib/answer_composition/multiple_guardrail/checker.rb
+++ b/lib/answer_composition/multiple_guardrail/checker.rb
@@ -11,7 +11,7 @@ module AnswerComposition::MultipleGuardrail
 
     MAX_TOKENS = 100
     SUPPORTED_MODELS = %i[claude_sonnet_4_0 claude_haiku_4_5].freeze
-    DEFAULT_MODEL = :claude_sonnet_4_0
+    DEFAULT_MODEL = :claude_haiku_4_5
 
     attr_reader :input, :llm_provider, :llm_prompt_name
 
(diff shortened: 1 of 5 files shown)


davidgisbey:
@kevindew i just noticed this and the failures. 

I think we will just need to do something like i've done in the structured response PR where we [pass the model in the chat options through to the response](https://github.com/alphagov/govuk-chat/pull/1050/changes/8d435cf8d1a5ba73899c35a0d3e67f7e7c0c87ac#diff-8034862fdbab1fe4e83d304d104408c2e2565a421f77ee8f4c40b5abb04bb9f8).

And update the guardrails helpers to default to haiku.

Give me a shout if you want me to pick this up since i know you're super busy.

kevindew:
Thanks @davidgisbey, appreciate the offer. I think I should be able to sort them. If you do have scope to sort them that'd be helpful.

kevindew (commented):
Nice one, thanks for all these fixes!

davidgisbey (approved):

PR #1067, opened 2026-04-27
[CHAT-502] Remove feedback buttons

chaecramb:
## What

Jira: https://gdsgovukagents.atlassian.net/jira/software/c/projects/CHAT/boards/269?label=Backend_Dev&selectedIssue=CHAT-502

Removes the Useful/Not useful answer feedback mechanism from Chat.

This removes answer feedback from:

- the public web interface, including the component, JavaScript behaviour, styles, route and controller action
- the admin interface, including question detail rendering and the usefulness filter on the questions list
- BigQuery exports
- the application model layer, including `AnswerFeedback`, its form object, associations, factories and specs
- the database, by adding a migration to drop the `answer_feedback` table

## Why

We no longer use answer feedback data, and the Chat API does not expose this functionality.

As we prepare to make the web version of Chat accessible to department users, we should avoid presenting a feedback mechanism that suggests this input is monitored or acted on when it is not.

## Notes

Existing historic answer feedback records from department testing are intentionally not retained.


spec/requests/admin/metrics_spec.rb
```diff
@@ -559,7 +559,7 @@
     end
 
     context "when period is last_7_days" do
-      it "returns data of answer feedback grouped by useful label by day" do
+      it "returns data of answer completeness grouped by completeness label by day" do
```
kevindew, line 562: Good spot - this must have been something old that was missed

kevindew (approved):
Looks great to me - really thorough job @chaecramb! It is always amazing how deeply embedded a feature can become into a system.

PR #1078, opened 2026-04-28
[CHAT-500] Add entry endpoint to handle users arriving from Signon

davidgisbey:
## Decription

We've got a bit of a problem at the moment with users arriving to Chat from Signon. Consider the following user groups:

- GOV.UK AI Team users who have access to the Admin Interface and want to use it for monitoring and evaluation purposes. Generally, these users will not have the `web-chat` permission and will be unable to access the Chat Interface.
- Departmental users who don't have access to the Admin Interface, but do have the `web-chat` permission and want to test Chat.

Both of these users will often, particularly the first time they arrive, use the link to the application in [Signon](https://signon.publishing.service.gov.uk/doorkeeper_applications/9615/edit).

The link currently points to the Admin UI, which is confusing for departmental users who may not know how to amend the url to get to the Chat Interface.

Ideally, we shouldn't have to either:

- Update it to point to the Chat Interface and have to let internal users know how to get to the Admin Interface, or
- Point it to the Admin Interface and have to let departmental users know how to get to the Chat Interface.

To get round this, we're going to add a new endpoint that we will link to from Signon. This endpoint will check the current_users permissions and:

- Redirect them to the Admin Interface if the have the `admin-area` permission
- Redirect them to the Chat Interface if they have the `web-chat` permission
- Prioritise the Admin Interface if they have both permissions
- Render the forbidden page if they have neither permission

The EntryController inherits from the BaseController which already handles user authentication, so we can be sure that the user is signed in when they hit this endpoint.

I've skipped the `authorise_web_user!` and `check_chat_web_access` before actions. The reason for this is:

- `authorise_web_user!`: We don't want this to run before the action as it would return a 403 for a user who has the `admin-area` permission but doesn't have the `web-chat` permission. As mentioned above, if a user doesn't have either permission then this endpoint will return a 403 forbidden response after checking permissions.

- `check_chat_web_access`: We only want to check that the Settings web chat access is enabled and render the downtime page if the user has the `web-chat` permission and doesn't have the `admin-area` permission. Due to this, i've called it within the conditional that is only reached after we've established that the user doesn't have the `admin-area` permission but does have the `web-chat` permission.

## Jira ticket

https://gdsgovukagents.atlassian.net/jira/software/c/projects/CHAT/boards/269?label=Backend_Dev&selectedIssue=CHAT-500


app/controllers/signon_entry_controller.rb
```diff
@@ -0,0 +1,14 @@
+class EntryController < BaseController
+  skip_before_action :authorise_web_user
+  skip_before_action :check_chat_web_access
+
+  def index
```
kevindew, line 5: Do we need to consider how a current_user is set (as I think we're skipping authorise_web_user) and the settings control of `ensure_signon_user_if_required`

davidgisbey: Unless i'm being daft this controller inherits from the BaseController so that will run before the action since we're not skipping it action. So they'd just get redirected to Signon. I did deploy the branch out and test it on Signon and that was the behaviour.

kevindew: So I was thinking that the method you're relying on in BaseController is `ensure_signon_user_if_required` which is:

```
  def ensure_signon_user_if_required
    return if Rails.configuration.available_without_signon_authentication

    authenticate_user!
  end
```

So I was thinking that if `Rails.configuration.available_without_signon_authentication` returns true, then you don't have a [current_user](https://github.com/alphagov/gds-sso/blob/f040fb9c8ab28cd59ebe683f31309a2601b78e25/lib/gds-sso/controller_methods.rb#L42-L44) and this would raise an exception.

It could be that it is easiest to just remove the conditional of `Rails.configuration.available_without_signon_authentication` since it has been a very long time since we considered setting it to false.

davidgisbey: Okay thanks. I've removed that 👍

config/routes.rb
```diff
@@ -20,6 +20,8 @@
                                     Mime::Type.lookup("text/html"),
                                     Mime::Type.lookup("application/json")] }
 
+  get "/entry", to: "entry#index", as: :entry
```
jackbot: Not sure if this was discussed but I'm wondering if "entry" is a bit vague if this route is only going to be used for entry from Signon? Would "/signon-entry" be more appropriate?

davidgisbey: Yea that's a good point. It does make it seem like it's the generic entry point when it's serving a more niche function.

I'll make that change.

davidgisbey:
@jackbot thanks for the review. I made that change.

jackbot (approved):
Nice!

PR #1079, opened 2026-04-28
[CHAT-523] Always rephrase question for Sonnet 4.5

davidgisbey:
**DO NOT MERGE until sign off from DS. Likely w/c 18/05**

## Description 

We want to update the QuestionRephraser to always rephrase the question, even if it is the users first question.

This updates the QuestionRephraser to do this for all models except Sonnet 4.0 as we don't have updated prompts for that model so it should continue to use the existing implementation.

We've also got a slight change due to always rephrasing, in that we need two user prompts. One for when there is history, and one for when there isn't.

I've added some basic tests to check that the correct user prompt is being used in both situations.

## Jira ticket 

https://gdsgovukagents.atlassian.net/jira/software/c/projects/CHAT/boards/269?label=Backend_Dev&selectedIssue=CHAT-523

### DIFF
--- a/lib/answer_composition/pipeline/question_rephraser.rb
+++ b/lib/answer_composition/pipeline/question_rephraser.rb
@@ -18,11 +18,11 @@ module AnswerComposition
       end
 
       def call
-        return if message_records.blank?
+        return if message_records.blank? && model_name == :claude_sonnet_4_0
 
         start_time = Clock.monotonic_time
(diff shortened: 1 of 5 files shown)


jackbot (approved):
LGTM :+1:

davidgisbey:
We're going to hold off on this until later next week due to a suspected increase of surge in traffic in the next week.

PR #1081, opened 2026-04-29
Move mock answer stubbing to system helper & remove unused helpers

davidgisbey:
## Description 

While spiking out rephrasing every question i noticed that there's some opportunities to move shared logic for stubbing mock answers into a helper.

Also we've got some unused helpers o i've removed those.


chaecramb (approved):
LGTM

PR #1107, opened 2026-05-13
[DO MOT MERGE] Revert "Temporarily increase API rate limits"

davidgisbey:
Reverts alphagov/govuk-chat#1093

## Description

We're no longer expecting a significant bump in users so we can revert this. Since we're expecting some increased traffic i'll keep an eye on it and increase it if we get anywhere near out rate limits.

We've not spiked above [1% of the current rate limit on grafana](https://grafana.eks.production.govuk.digital/d/govuk-chat-technical/gov-uk-chat-technical?orgId=1&from=now-2d&to=now&timezone=browser&refresh=15m) at the moment which would be 5% with these changes.


davidgisbey:
Closing because we've been given a street to keep these elevated.

PR #1124, opened 2026-05-21
[CHAT-645] Ensure Faithfulness evaluation sends all retrieved sources in the correct format

davidgisbey:
## Description

There are a couple of issues Nick identified in our Faithfulness evaluation:

1. We're not sending all the chunks that we sent to the LLM to generate an answer. Currently, we're only sending the chunks from used sources.
2. We're not sending the chunks to the LLM in the correct format. We're using the `#plain_content` method. This strips out the html from the content, including the inline links.

I've updated the evaluation to send all the sources that we send to the structured answer comp llm call and updated the chunk content to [mirror the implementation in the evaluation repo](https://github.com/alphagov/govuk-chat-evaluation/blob/08894080765b0b4c21451adc3c13d3370359977c/govuk_chat_evaluation/rag_answers/data_models/input.py#L15-L22).

## Jira ticket

https://gdsgovukagents.atlassian.net/jira/software/c/projects/CHAT/boards/269?label=Backend_Dev&selectedIssue=CHAT-645

### DIFF
--- a/lib/auto_evaluation/faithfulness.rb
+++ b/lib/auto_evaluation/faithfulness.rb
@@ -49,16 +49,12 @@ private
   end
 
   def retrieval_context
-    used_sources.map(&:plain_content).join("\n\n")
+    answer.sources.map(&:plain_content).join("\n\n")
   end
 
   def calculate_score(verdicts)
     faithful_count = verdicts.count { |verdict| verdict["verdict"].strip.downcase != "no" }
     faithful_count.to_d / verdicts.count
-  end
-
-  def used_sources
-    answer.sources.select(&:used)
   end
 
   def build_error_result(error_message)
--- a/lib/auto_evaluation/faithfulness.rb
+++ b/lib/auto_evaluation/faithfulness.rb
@@ -49,7 +49,7 @@ private
   end
 
   def retrieval_context
-    answer.sources.map(&:plain_content).join("\n\n")
+    answer.sources.map { |source| format_chunk_for_evaluation(source.chunk) }.join("\n")
   end
 
   def calculate_score(verdicts)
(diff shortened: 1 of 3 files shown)


lib/auto_evaluation/faithfulness.rb
```diff
@@ -79,4 +75,16 @@ def build_result_with_score(score, reason)
       metrics:,
     )
   end
+
+  def format_chunk_for_evaluation(chunk)
```
langenk, line 79: The format you implemented here is actually the one we specifically use for context_relevancy. Faithfulness in govuk-chat-evaluation gets the slightly simpler one (to_flattened_string() not to_flattened_context_content()) https://github.com/alphagov/govuk-chat-evaluation/blob/08894080765b0b4c21451adc3c13d3370359977c/govuk_chat_evaluation/rag_answers/data_models/input.py#L15-L22

lib/auto_evaluation/faithfulness.rb
```diff
@@ -49,18 +49,14 @@ def answer_message
   end
 
   def retrieval_context
-    used_sources.map(&:plain_content).join("\n\n")
+    answer.sources.map { |source| format_chunk_for_evaluation(source.chunk) }.join("\n")
```
langenk, line 52: Using full answer.sources looks good to me

lib/auto_evaluation/faithfulness.rb
davidgisbey: Ah sorry @langenk. I've updated that.

langenk (approved):
Great, looks good to me now. :)

kevindew (approved):

PR #1134, opened 2026-05-26
Revert "[CHAT-523] Always rephrase question for Sonnet 4.5"

kevindew:
Reverts alphagov/govuk-chat#1079

We're reverting this because we have experienced a number of reasoning leaks where additional context is added to the rephrase that wasn't provided by the user.


chaecramb (approved):
Looks good.

PR #1151, opened 2026-06-08
[CHAT-560] Model switch cleanup

davidgisbey:
## Description 

This PR removes Sonnet 4.0 and Sonnet 4.6 support from the codebase. Sonnet 4.0 because we're no longer using it and Sonnet 4.6 because we ended up using 4.5 instead.

Sonnet 4.6 was simple to remove. I simply removed references to it from the stubs and remove it from the BedrockModels and healthchecks.

Sonnet 4.0 is a little more involved, but still pretty straightforward. I:

- Removed all references to it from the AnswerCompositionPipeline (supported models, strict mode logic and specs)
- Removed references to it from helper methods and some other references in tests
- Remove it from BedrockModels class and various related tests (healthchecks etc.)

Where possible i've either used a contrived model or used the `BedrockModels::MODEL_IDS` constant so that we don't need to update these tests each time a model switch occurs/one is deprecated.

## Jira ticket

https://gdsgovukagents.atlassian.net/jira/software/c/projects/CHAT/boards/269?label=Backend_Dev&selectedIssue=CHAT-560


jackbot (approved):
NIce one

PR #1167, opened 2026-06-15
[CHAT-705] Update faithfulness to penalise idk verdicts

davidgisbey:
## Description 

We want to update update the faithfulness metric to deviate from the current implementation of treating 'idk' verdicts as faithful and instead treat them as 'no' values during score calculation.

As part of this the reasoning prompt has been slightly tweaked to indicate that ambiguous and contradictions are being passed with the score for analysis. This was merged [here](https://github.com/alphagov/govuk_chat_private/pull/136).

A couple of follow up PRs will be opened to clean up the old prompt after this is merged.

## Jira ticket

https://gdsgovukagents.atlassian.net/jira/software/c/projects/CHAT/boards/269?label=Backend_Dev%2Cbackend_dev&selectedIssue=CHAT-705

### DIFF
--- a/lib/auto_evaluation/faithfulness/reason_generator.rb
+++ b/lib/auto_evaluation/faithfulness/reason_generator.rb
@@ -24,19 +24,23 @@ module AutoEvaluation
 
     def user_message
       sprintf(
-        llm_prompts.fetch(:user_prompt),
+        llm_prompts.fetch(:new_user_prompt),
         score:,
-        contradictions:,
+        unfaithful_claims:,
       )
     end
 
     def tool
-      llm_prompts.fetch(:tool_spec)
+      llm_prompts.fetch(:new_tool_spec)
     end
 
-    def contradictions
-      verdicts.select { |verdict| verdict["verdict"].strip.downcase == "no" }
-              .map { |verdict| verdict["reason"] }
+    def unfaithful_claims
+      verdicts.filter_map do |verdict|
+        status = verdict["verdict"].strip.downcase
+        next if status == "yes"
+
+        status == "idk" ? "(Ambiguous) #{verdict['reason']}" : "(Contradiction) #{verdict['reason']}"
+      end
     end
   end
 end
--- a/lib/auto_evaluation/faithfulness.rb
+++ b/lib/auto_evaluation/faithfulness.rb
@@ -24,7 +24,7 @@ class AutoEvaluation::Faithfulness
 
     return build_error_result("No verdicts were generated for the extracted claims.") if verdicts.empty?
 
-    if verdicts.none? { |verdict| verdict["verdict"].strip.downcase == "no" }
+    if verdicts.all? { |verdict| verdict["verdict"].strip.downcase == "yes" }
       return build_result_with_score(1.0, "The response is fully supported by the retrieval context.")
     end
 
@@ -53,7 +53,7 @@ private
   end
 
   def calculate_score(verdicts)
-    faithful_count = verdicts.count { |verdict| verdict["verdict"].strip.downcase != "no" }
+    faithful_count = verdicts.count { |verdict| verdict["verdict"].strip.downcase == "yes" }
     faithful_count.to_d / verdicts.count
   end
 
--- a/spec/lib/auto_evaluation/faithfulness/reason_generator_spec.rb
+++ b/spec/lib/auto_evaluation/faithfulness/reason_generator_spec.rb
@@ -1,26 +1,31 @@
 RSpec.describe AutoEvaluation::Faithfulness::ReasonGenerator, :aws_credentials_stubbed do
   describe ".call" do
-    let(:score) { 0.5 }
+    let(:score) { 0.0 }
     let(:verdicts) do
       [
         { "verdict" => "no", "reason" => "The retrieval context states Einstein won in 1921, not 1968." },
-        { "verdict" => "yes" },
+        { "verdict" => "idk", "reason" => "The retrieval context does not explicitly confirm or deny that Einstein won the Nobel Prize for the photoelectric effect." },
       ]
     end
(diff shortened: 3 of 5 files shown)


lib/auto_evaluation/faithfulness/reason_generator.rb
```diff
@@ -35,7 +35,7 @@ def tool
     end
 
     def contradictions
-      verdicts.select { |verdict| verdict["verdict"].strip.downcase == "no" }
+      verdicts.reject { |verdict| verdict["verdict"].strip.downcase == "yes" }
```
davidgisbey: @ciaran-ryan-gds i've got a question around the reason LLM call we make. With the current implementation we're not passing idk verdicts to the [reasoning prompt](https://github.com/alphagov/govuk_chat_private/blob/c52bce8e92e1965d15af01c677f61d25f47c24d1/config/llm_prompts/auto_evaluation/faithfulness.yml#L147-L168).

This means the score could theoretically be 0 with all idk no verdicts, but no contradictions would be passed.

I think we can either:

1. pass the idk verdict reasons in the contradictions (even though they're not necessarily contradictions
2. update the prompt to have a section for idk verdicts

So from:

```
Faithfulness Score:
%{score}

Contradictions:
%{contradictions}
```

to something like:

```
Faithfulness Score:
%{score}

Contradictions:
%{contradictions}

Ambiguous verdicts (counted as unfaithful):
%{ambiguous_verdicts}
```

@ciaran-ryan-gds let me know if this doesn't make sense and what your preference is.

davidgisbey: oh actually i've just had a look what [deepeval do](https://github.com/confident-ai/deepeval/blob/a8c764ce921ffd29fa3379733245aa17017e07aa/deepeval/metrics/faithfulness/faithfulness.py#L183C15-L183C71).

They just prepend the verdicts with `(ambiguous)` so we can just do that unless you disagree.

ciaran-ryan-gds: we definitely want reasoning for these 'idk' verdicts alright. 

Just checking, when they prepend verdicts with 'ambiguous', are they still being presented to the llm-as-a-judge as 'contradiction'? So the judge is being asked something like 

> Please explain why these were unfaithful:
> 
> Contradictions:
> (Contradiction) xyz
> (Ambiguous) abc

If this is the case, it might be confusing to the llm to put the 'ambiguous' verdicts under a "contradictions" header, since they're not strictly contradictions. 

Maybe we could define the umbrella grouping "unfaithful claims" and then prep end them with Condtradiction and Ambiguous? So the llm as a judge sees something more like.

> Please explain why these were unfaithful:
> 
> Unfaithful Claims:
> (Contradiction) xyz
> (Ambiguous) abc

But let me know if I'm misunderstanding how the claims are currently presented to the judge. Your initial suggestion of just having them completely separate also works for me as it should be clear to the LLM!

davidgisbey: No that's exactly right. Here's the [code that builds the contradictions](https://github.com/confident-ai/deepeval/blob/a8c764ce921ffd29fa3379733245aa17017e07aa/deepeval/metrics/faithfulness/faithfulness.py#L175-L184) and they're passed into the prompt [here](https://github.com/confident-ai/deepeval/blob/a8c764ce921ffd29fa3379733245aa17017e07aa/deepeval/metrics/faithfulness/template.py#L220-L221).

So i'll update the prompt from `Contradictions` to `Unfaithful claims` and prepend Contradiction or Ambiguous based on verdict 👍

ciaran-ryan-gds (approved):
lgtm, thanks for the quick turnaround. Confirmed we only count 'yes' towards the faithfulness score now, and generate reasons for idk verdicts

jackbot (approved):

lib/auto_evaluation/faithfulness/reason_generator.rb
```diff
@@ -24,19 +24,23 @@ def llm_prompts
 
     def user_message
       sprintf(
-        llm_prompts.fetch(:user_prompt),
+        llm_prompts.fetch(:new_user_prompt),
```
kevindew, line 27: I'm expecting that this naming is only temporary? Just aware that prefixing anything with `new_` is something that tends to age badly in software variables.

Can't imagine it's worth changing if we're going to shortly after remove them, though - something to bear in mind in future though (I'd suggest something like `idk_penalised_user_prompt`)

lib/auto_evaluation/faithfulness/reason_generator.rb
```diff
@@ -24,19 +24,23 @@ def llm_prompts
...
     def tool
-      llm_prompts.fetch(:tool_spec)
+      llm_prompts.fetch(:new_tool_spec)
     end
 
-    def contradictions
-      verdicts.select { |verdict| verdict["verdict"].strip.downcase == "no" }
-              .map { |verdict| verdict["reason"] }
+    def unfaithful_claims
+      verdicts.map { |verdict|
```
kevindew: minor, but if you use a filter_map here you won't need a compact and it'll be a bit neater with a do / end block

kevindew (commented):
Looks great, just a minor couple of comments really

lib/auto_evaluation/faithfulness/reason_generator.rb
davidgisbey: Totally fair. I've already got a PR open to update the prompts to be have duplicate names and will remove it shortly after.

lib/auto_evaluation/faithfulness/reason_generator.rb
davidgisbey: yea good shout i've sorted that.

PR #1194, opened 2026-07-01
[CHAT-824] Address results of AI security audit

jackbot:
https://gdsgovukagents.atlassian.net/browse/CHAT-824

See individual commits for details. The report is attached to the Jira ticket.

### DIFF
--- a/lib/api/rate_limit/middleware.rb
+++ b/lib/api/rate_limit/middleware.rb
@@ -6,6 +6,7 @@ class Api::RateLimit::Middleware
     Api::RateLimit::GOVUK_END_USER_WRITE_THROTTLE_NAME => "Govuk-End-User-Id-Write",
   }.freeze
   SLACK_NOTIFICATION_THRESHOLD_PERCENTAGE = 75
+  SLACK_NOTIFICATION_PERIOD = 15.minutes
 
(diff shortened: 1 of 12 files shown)


lib/api/rate_limit/middleware.rb
```diff
@@ -47,9 +47,14 @@ def call(env)
         )
 
         if percentage_used >= SLACK_NOTIFICATION_THRESHOLD_PERCENTAGE
-          NotifySlackApiUserRateLimitWarningJob.perform_later(
-            user_name, percentage_used, "read"
-          )
+          cache_key = "slack_rate_limit_warning:#{user_name}:read:#{epoch_time / period}"
```
kevindew: I think we'd want to go longer for the cache key here as for most of the alerts, we'd get a message every minute? I'd imagine we'd want to go for a fixed time of 15 mins?

kevindew (commented):
Looks great, just one change suggested. Thanks for picking this up.

lib/api/rate_limit/middleware.rb
jackbot: Good call, updated it to 15 mins.

kevindew (approved):
Nice one

PR #1199, opened 2026-07-07
Add QuestionRephraser to RAG answer generator task

kevindew:
Previously the question rephraser only ran when a conversation had multiple messages, making it unnecessary for testing RAG. Now however we use it to rephrase even the first question in a conversation, therefore in order to test a RAG answer correctly we need to include it.


jackbot (approved):

kevindew:
I'm closing this as my impression that this was a simple bug fix was incorrect. Instead while the rephraser can impact the RAG responses it is not an agreeable position with our Data Scientists that this should be part of testing RAG answers. 

I expect a need for this type of thing to arrive again later but quite what it is tested is subject to a later debate.

PR #1210, opened 2026-07-15
Add rake task to output document types

kevindew:
Despite previous efforts [1], it's still a struggle to communicate what GOV.UK Content is indexed with non-technical folk.

Following a request for this information on Slack I quickly ran some code to get the data out and have now committed it to the repo so it can be reused.

[1]: https://github.com/alphagov/govuk-chat/pull/701


jackbot (approved):
Nice

PR #1267, opened 2026-08-27
[CHAT-968] Add AnswerFeedback API endpoints

davidgisbey:
## Description 

This largely re-implements the feedback mechanism we had in place and removed in this [PR](https://github.com/alphagov/govuk-chat/pull/1067). There are some notable exceptions. We've gone with a `reason` enum instead of a `useful` boolean to capture the response. Also i've made a couple of small tweaks which i'll leave comments on.

This PR:

1. re-adds the table via a migration
2. adds the model & factory
3. adds the POST feedback endpoint (also re-adds the form for validation)
4. extends the answer blueprint to also include feedback so that it's sent as part of the GET conversation and messages response payload
5. Re-adds it to the big query export

It should be noted that there won't be any issues with clashing values in the BigQuery tables because they integration and production tables were dropped prior to us going live. 

## Out of scope

I've not looked into reworking the way we do rate limiting as part of this PR. I'll look at that in a follow up bit of work.

## Jira ticket

https://gdsgovukagents.atlassian.net/jira/software/c/projects/CHAT/boards/269?label=Backend_Dev&selectedIssue=CHAT-968

### DIFF
--- a/app/blueprints/answer_blueprint.rb
+++ b/app/blueprints/answer_blueprint.rb
@@ -11,4 +11,18 @@ class AnswerBlueprint < Blueprinter::Base
       { url: source[:href], title: source[:title] }
     end
   end
+
+  field :feedback, if: ->(_field_name, answer, _options) { answer.feedback.present? } do |answer, _options|
+    {
+      reaction: answer.feedback.reaction,
+      created_at: answer.feedback.created_at.iso8601,
+    }
+  end
+
+  field :feedback_url, if: ->(_field_name, answer, _options) { answer.feedback.blank? } do |answer, _options|
+    Rails.application.routes.url_helpers.api_v1_answer_feedback_path(
+      answer.question.conversation_id,
+      answer.id,
+    )
+  end
 end
--- a/app/controllers/api/v1/conversations_controller.rb
+++ b/app/controllers/api/v1/conversations_controller.rb
@@ -1,6 +1,8 @@
 class Api::V1::ConversationsController < Api::BaseController
   before_action { authorise_user!(SignonUser::Permissions::CONVERSATION_API) }
-  before_action :find_conversation, only: %i[show update answer questions]
+  before_action :find_conversation, only: %i[show update answer questions answer_feedback]
+
+  FEEDBACK_ALREADY_PROVIDED_MESSAGE = "Feedback already provided for this answer".freeze
 
   def create
     conversation = Conversation.new(
@@ -110,19 +112,40 @@ class Api::V1::ConversationsController < Api::BaseController
     render(json:, status: :ok)
   end
 
+  def answer_feedback
+    answer = @conversation.answers.includes(:feedback).find(params[:answer_id])
+    return render_feedback_already_provided if answer.feedback.present?
+
+    form = Form::CreateAnswerFeedback.new(answer_feedback_params.merge(answer:))
+
+    unless form.valid?
+      return render json: ValidationErrorBlueprint.render(
+        errors: form.errors.messages,
+      ), status: :unprocessable_content
+    end
+
+    form.submit
+
+    render json: {}, status: :created
+  rescue ActiveRecord::RecordNotUnique
+    render_feedback_already_provided
+  end
+
 private
 
   def find_conversation
-    where = { signon_user_id: current_user.id, source: :api, end_user_id: end_user_id_header }
-
     @conversation = Conversation
                     .active
-                    .where(where)
+                    .where(signon_user_id: current_user.id, source: :api, end_user_id: end_user_id_header)
                     .find(params[:conversation_id])
   end
 
   def question_params
     params.permit(:user_question)
+  end
+
+  def answer_feedback_params
+    params.permit(:reaction)
   end
 
   def answer_path(question)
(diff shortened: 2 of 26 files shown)


app/controllers/api/v1/conversations_controller.rb
```diff
@@ -110,21 +112,42 @@ def questions
...
+    unless form.valid?
+      return render json: ValidationErrorBlueprint.render(
+        errors: form.errors.messages,
+      ), status: :unprocessable_content
+    end
+
+    form.submit
+
+    render json: {}, status: :created
+  rescue ActiveRecord::RecordNotUnique
```
davidgisbey, line 130: this is slightly different to the original implementation. In that we checked for feedback [within the form itself](https://github.com/alphagov/govuk-chat/pull/1067/changes#diff-f2c55e710929102e645179ae1c0323ee58aa445f9d130561a97db25f7e2027b6L9) and returned a 422 if it already had feedback. 

I think the more conventional approach is to return a 409. I've chosen to remove the validation from the form object and let the error bubble if feedback is persisted between the check above on line 117 and the time we attempt to persist the object within the form.

That way we can just catch it and return a 409.

app/blueprints/answer_blueprint.rb
```diff
@@ -11,4 +11,18 @@ class AnswerBlueprint < Blueprinter::Base
...
   end
+
+  field :feedback, if: ->(_field_name, answer, _options) { answer.feedback.present? } do |answer, _options|
+    {
+      reaction: answer.feedback.reaction,
+      created_at: answer.feedback.created_at.iso8601,
+    }
+  end
+
+  field :feedback_url, if: ->(_field_name, answer, _options) { answer.feedback.blank? } do |answer, _options|
```
davidgisbey, line 22: i chose to have conditional fields rather than always having the 3 fields and nulling them based on state. 

ie.

```

field :feedback do |answer, _options|
  {
    created_at: answer.feedback&.created_at&.iso8601,
    reaction: answer.feedback&.reaction,
    url: answer.feedback.present? ? nil : Rails.application.routes.url_helpers.api_v1_answer_feedback_path(answer.question.conversation_id, answer.id),
  }
end

```
Happy to discuss if people disagree

db/migrate/20260827090000_create_answer_feedback.rb
```diff
@@ -0,0 +1,14 @@
+class CreateAnswerFeedback < ActiveRecord::Migration[8.1]
+  def change
+    create_enum :answer_feedback_reaction, %w[positive negative]
+
+    create_table :answer_feedback, id: :uuid do |t|
+      t.references :answer, type: :uuid, null: false, index: { unique: true }, foreign_key: { on_delete: :cascade }
+      t.enum :reaction, null: false, enum_type: "answer_feedback_reaction"
+
+      t.timestamps
```
jackbot, line 9: Do we need the `update_at` field? There can only ever be a maximum 1 record per answer, and it's not possible to change it, so it feels like a `created_at` would suffice?

app/controllers/api/v1/conversations_controller.rb
```diff
@@ -110,21 +112,42 @@ def questions
...
+
 private
 
   def find_conversation
-    where = { signon_user_id: current_user.id, source: :api, end_user_id: end_user_id_header }
-
     @conversation = Conversation
                     .active
-                    .where(where)
+                    .where(signon_user_id: current_user.id, source: :api, end_user_id: end_user_id_header)
```
jackbot, line 139: Any reason this was changed out of interest?

jackbot (commented):
Looks great, nice work. One tiny comment.

db/migrate/20260827090000_create_answer_feedback.rb
davidgisbey: Yea it shouldn't ever really be needed. I've not actually seen only using `created_at` before in a Rails project so i'd just assumed (maybe incorrectly) that it was convention to add both. It sounds like maybe that's not the case though?

In terms of this codebase, we've actually got other examples where updated_at shouldn't ever be touched (bigquery_exports, settings_audits & evaluation runs etc.), but still have one.

So i guess we can either just add it here or remove it from the other places. I'd lean towards the former, but i'm not overly attached to it if you if disagree.

app/controllers/api/v1/conversations_controller.rb
davidgisbey: I actually moved the code to a base controller as i'd forgotten we'd originally implemented the feedback endpoint within the convo controller. When i reverted i forgot to revert this.

Happy to revert if if you'd prefer, but i think it's the only place we pass logic into a where clause so it's probs slightly more inline with the rest of the repo.

db/migrate/20260827090000_create_answer_feedback.rb
jackbot: Nah sounds good, let's leave it in.

app/controllers/api/v1/conversations_controller.rb
jackbot: All good, thanks for explaining.

jackbot (dismissed):
Nice one :+1:

docs/api_openapi_specification.yml
```diff
@@ -23,7 +23,7 @@ info:
     | API User  | 900 RPM             | 10,800 RPM          |
     | End User  | 15 RPM              | 180 RPM            |
 
-  version: "0.5.0"
```
davidgisbey, line 26: Sorry @jackbot i just realised i didn't bump the version of the spec so i've just pushed that up which will have dismissed your review.

jackbot (approved):

PR #1272, opened 2026-09-01
[CHAT-968] Update rate limiting strategy

davidgisbey:
## Description 

This PR updates our rate limiting strategy for the API. Previously, we used HTTP methods to implement read and write rate limits.

Instead we now use conversation write ad default limits. The conversation write covers interactions that trigger model inference as part of the requests since those are the expensive requests. All other requests now fall under the default rate limits.

The motivation for this change was that we added feedback endpoints to the API and if a user rapidly gave feedback there was a slightly concern they could hit the write rate limit and stop answer generation. 

The feedback endpoints don't invoke models and while they do interact with the DB, our Postgres instance has been load tested sufficiently for it not be an attack vector of concern.

In terms of implementation the Jira ticket calls out that the App team shouldn't be affected by this change. Due to this
constraint we've retained the same header names (read and write) which could be slightly misleading.

I did consider sending additional headers (`conversation-write` and `default`) so that we could do something like:

1. send duplicate headers
2. allow the app team to migrate to using the new headers
3. removing the older headers

In reality we're going to be moving to Chat V2 which will likely move to new headers so for now it makes sense not to cause any integration issues and retain the existing header structure.

Our codebase has been updated to reflect the change (naming of methods etc.) and the OpenAPI spec has documented the changes to the rate limits though.

## Jira ticket

https://gdsgovukagents.atlassian.net/jira/software/c/projects/CHAT/boards/269?label=Backend_Dev&selectedIssue=CHAT-968

### DIFF
--- a/lib/api/rate_limit.rb
+++ b/lib/api/rate_limit.rb
@@ -1,6 +1,30 @@
 class Api::RateLimit
-  GOVUK_API_USER_READ_THROTTLE_NAME = "read requests to Conversations API with token".freeze
-  GOVUK_API_USER_WRITE_THROTTLE_NAME = "write method requests to Conversations API with token".freeze
-  GOVUK_END_USER_READ_THROTTLE_NAME = "read requests to Conversations API with user id".freeze
-  GOVUK_END_USER_WRITE_THROTTLE_NAME = "write requests to Conversations API with user id".freeze
+  GOVUK_API_USER_DEFAULT_THROTTLE_NAME = "default requests to Conversations API with token".freeze
+  GOVUK_API_USER_CONVERSATION_WRITE_THROTTLE_NAME = "conversation write requests to Conversations API with token".freeze
+  GOVUK_END_USER_DEFAULT_THROTTLE_NAME = "default requests to Conversations API with user id".freeze
+  GOVUK_END_USER_CONVERSATION_WRITE_THROTTLE_NAME = "conversation write requests to Conversations API with user id".freeze
+
+  CONVERSATION_API_PATH_REGEX = /^\/api\/v\d+\/conversation/
+  CONVERSATION_API_CREATE_PATH_REGEX = /^\/api\/v\d+\/conversation$/
+  CONVERSATION_API_UPDATE_PATH_REGEX = /^\/api\/v\d+\/conversation\/[^\/]+$/
+
+  def self.conversation_write?(request)
+    create_conversation?(request) || update_conversation?(request)
+  end
+
+  def self.default?(request)
+    conversation_api_request?(request) && !conversation_write?(request)
+  end
+
+  def self.conversation_api_request?(request)
+    request.path.match?(CONVERSATION_API_PATH_REGEX)
+  end
+
+  def self.create_conversation?(request)
+    request.post? && request.path.match?(CONVERSATION_API_CREATE_PATH_REGEX)
+  end
+
+  def self.update_conversation?(request)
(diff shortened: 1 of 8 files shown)


lib/api/rate_limit.rb
```diff
@@ -1,6 +1,27 @@
...
+  GOVUK_END_USER_DEFAULT_THROTTLE_NAME = "default requests to Conversations API with user id".freeze
+  GOVUK_END_USER_CONVERSATION_WRITE_THROTTLE_NAME = "conversation write requests to Conversations API with user id".freeze
+
+  CONVERSATION_API_PATH_REGEX = /^\/api\/v\d+\/conversation/
+  # Captures an additional optional path segment after /conversation to match
+  # creating a conversation and updating one.
+  CONVERSATION_WRITE_PATH_REGEX = /^\/api\/v\d+\/conversation(\/[^\/]+)?$/
+
+  def self.conversation_write?(request)
+    !read_method?(request) && request.path.match?(CONVERSATION_WRITE_PATH_REGEX)
```
jackbot: Could this be done with `!read_method?(request) && request.path.match?(CONVERSATION_API_PATH_REGEX)`?

Because a "conversation write" request is just a non-read request on a path beginning with `/api/v1/conversation`. The `CONVERSATION_API_PATH_REGEX` regex should cover the path match. Or am I missing something?

spec/requests/api/middleware_spec.rb
```diff
@@ -99,52 +99,52 @@
       end
     end
 
-    describe "/api/v1/conversations write rate limits" do
-      it "treats non read requests as write requests with rate limits" do
-        %i[post put patch delete].each do |method|
-          public_send(method, "/api/v1/conversations/404")
+    describe "conversation write rate limits" do
+      it "applies to writes that create or update a conversation" do
+        %i[post patch delete].each do |method|
```
jackbot: How come `put` was removed here? From the routes file we only actually support `PUT` for `/api/v1/conversation/:id`. Which makes me wonder why we had `%i[post put patch delete]` here before, since we don't support most of those routes.

davidgisbey: ah sorry claude code pulled that out and i missed it. I'll revert that.

lib/api/rate_limit.rb
davidgisbey: Yea i think there's a bit of context missing that i thought i'd put in the commit message but i didn't.

The issue with just using the current regex is that it also matched the POST feedback endpoint `/api/v1/:conversation_id/answer/:answer_id/feedback`.

And we only want to put endpoints that invoke models into that reduced rate limit since those are expensive. Giving feedback only hits the DB so it falls into the `default` bucket.

spec/requests/api/middleware_spec.rb
davidgisbey: Yea i guess this implementation could be far more specific. Let me have a quick think about changing it to specifically match on those endpoints.

davidgisbey:
Thanks for the review @jackbot. I've tweaked the first commit so it specifically only matches against the create and update conversation endpoints  since those are the only endpoints that are going to go into the rate limits before we deprecate V1.

lib/api/rate_limit.rb
jackbot: Ahh yeah I didn't clock that the feedback URL would be included. Cool, makes sense.

jackbot (approved):
Nice one, I think it's much clearer now :+1:

PR #1276, opened 2026-09-02
[CHAT-904] Add answer analysis request type

davidgisbey:
## Description

This PR adds a new evaluation type to the answer analysis workflow. This captures the request type (intent). The implementation of this closely mirrors the topics one. The structure of the response we want to capture is pretty similar:

- There is a primary type and an optional secondary type
- It only runs once directly after an answer is generated
- It's association with answer in the same has_one/belongs_to
- We export it to BigQuery
- It was a similar rake task that can be used to generate the object
- It'll rendered in the admin UI in essentially the same way as topics

This PR adds all of the above with the exception of the Admin UI  work. That'll be handled in a follow-up PR.

## Example object

image

## Jira ticket

https://gdsgovukagents.atlassian.net/jira/software/c/projects/CHAT/boards/269?label=Backend_Dev&selectedIssue=CHAT-904

### DIFF
--- a/app/jobs/answer_analysis/tag_request_types_job.rb
+++ b/app/jobs/answer_analysis/tag_request_types_job.rb
@@ -1,0 +1,31 @@
+module AnswerAnalysis
+  class TagRequestTypesJob < BaseJob
+    def perform(answer_id)
+      answer = Answer.includes(:request_types, question: :conversation).find_by(id: answer_id)
+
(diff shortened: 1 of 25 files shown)


jackbot (commented):
All looks good to me. I'll hold off on actually approving it until the prompt has been merged into the private repo though.

chaecramb (approved):
LGTM

PR #1287, opened 2026-09-09
Add a rake task to backfill request types

davidgisbey:
## Description 

There are ~62,000 answers eligible for request type analysis that
predate the tagging, and each needs its own LLM call.

This adds a temporary take task that will be removed after we've backfilled
request type.

The task deliberately doesn't enqueue TagRequestTypesJob. That job
shares a 300 per hour auto-evaluation quota with every other analysis
job, so a backfill routed through it would exhaust the quota within
seconds and silently drop the work, stopping live auto-evaluation for
the duration. The logic for persisting a tagger result is duplicated
from the job rather than shared, as this task will be deleted once the
backfill is done.

Answers are read in batches of 1,000 and only their id and question are
selected, to avoid loading the whole backlog, and the large llm_responses
and metrics columns, into memory at once. Batching by id is safe here
even though tagging an answer removes it from the scope, as the rows
that stop matching are always behind the cursor.

Each batch is then tagged by a pool of 10 threads, which brings the
backfill down from around 36 hours to under 4. Bedrock's limits for
gpt-oss-120b are 10k requests and 100M tokens a minute, so this uses a
few percent of the available quota and leaves plenty of head room for
live auto-evaluation, which is capped at 300 evaluations an hour.

Interrupting the task is safe, as answers that already have request
types are excluded, so a later run picks up where it left off. Note that
answers tagged with an error status aren't retried by a later run.

## Testing on integration 

I ran it on 200 records on integration and it took ~35 seconds 1.19% usage of the rate limit.

```
govuk-chat(prod):001> a = AnswerAnalysis::RequestTypes.order(created_at: :desc).limit(200)

govuk-chat(prod):003> a.map(&:status).tally
=> {"success" => 200}
govuk-chat(prod):004> a.map(&:primary_request_type).tally
=>
{"factual_lookup" => 26,
 "verify_identity" => 34,
 "explore_topic" => 12,
 "update_details" => 9,
 "government_issue" => 12,
 "retrieve_own_information" => 26,
 "do_task" => 40,
 "rules_compliance" => 22,
 "eligibility" => 8,
 "track_application_status" => 4,
 "unclassifiable" => 2,
 "rule_or_service_change" => 1,
 "unknown_new_intent" => 1,
 "human_handoff" => 3}
govuk-chat(prod):005>

```

and with batches of 10 to show it works over multiple batches

```
not starting Prometheus metrics server: address already in use
Backfilling request types for 61264 answers with a concurrency of 10
(10 / 61264)
(20 / 61264)
(30 / 61264)

=>
["success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success",
 "success"]
govuk-chat(prod):005> AnswerAnalysis::RequestTypes.order(created_at: :desc).limit(30).map(&:primary_request_type).tally
=>
{"government_issue" => 5,
 "retrieve_own_information" => 6,
 "factual_lookup" => 2,
 "unknown_new_intent" => 1,
 "rules_compliance" => 1,
 "do_task" => 5,
 "track_application_status" => 1,
 "verify_identity" => 4,
 "explore_topic" => 1,
 "update_details" => 3,
 "unclassifiable" => 1}
```

## Jira ticket

https://gdsgovukagents.atlassian.net/jira/software/c/projects/CHAT/boards/269?label=Backend_Dev

### DIFF
--- a/lib/bigquery/individual_export.rb
+++ b/lib/bigquery/individual_export.rb
@@ -2,6 +2,8 @@ require "google/cloud/bigquery"
 
 module Bigquery
   class IndividualExport
+    BATCH_SIZE = 1_000
+
     Result = Data.define(:tempfile, :count)
 
     def self.remove_nil_values(json)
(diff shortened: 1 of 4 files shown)


jackbot (approved):
Nice one

